PR #1, opened 2017-06-20
First commit

sohamm17:


htuch:
Can you squash down all these commits to a single commit and add a meaningful commit message for reviewers? Thanks.

htuch:
You will need to do a `git push origin -f`, this whole PR history should just become one commit I think.

PR #2, opened 2017-06-20
Performance Benchmarking for Envoy

sohamm17:
Bencharking is being done by running a top-level Python script. The script creates a VM in Google Cloud and performs benchmarking inside the VM. In the end, it deletes the VM.
 
For now, it runs Nginx server and Envoy with a loopback configuration. `h2load` is used to generate load directly on Nginx server and through Envoy proxy. Result is saved as a JSON file.


htuch:
@mattklein123 I've already reviewed this, looks like a good starting point for orchestrating basic load testing on GCE. Do you have any thoughts?

mattklein123:
I don't have time to go through this in detail. One thing I would point out for future work is that for horse race benchmarks there are some envoy features we might want to turn off (or profile both ways). I can discuss those separately. If you have looked through I would just go ahead and merge and we can iterate as things evolve.

htuch:
Yeah, we recognized the need for config customization, hence the use of jinja2 templating. https://github.com/lyft/envoy-perf/issues/3 is the tracking issue for further customization, when we start producing real number we should make sure we followup on this. I'll merge this based on your guidance above.

templates/nginx.template.conf
```diff
@@ -0,0 +1,98 @@
+#user sohamcodes;
```
jamessynge: I suspect this file and the other template need some cleanup. For example, they have big commented out sections, and have comments that probably came from elsewhere but may not apply here. And it would be good for the files to have comments at the top describing why they exist/what they're used for.

Readme.md
```diff
@@ -0,0 +1,60 @@
+**Pre-requisite:**
+
+1. python2
```
jamessynge: I'd add lyft/envoy to the list, even though it seems obvious. And are you using nginx as the backend for the load? If so, it belongs in this list.

Readme.md
```diff
@@ -0,0 +1,60 @@
...
+2. gcloud
+
+Follow these set-up before running the benchmarking script.
```
jamessynge: These sound like steps that could be better described as "Edit <some-template-script> and set the values of these variables (e.g. ENVOY_BINARY) to the appropriate values."

Readme.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. Keep your envoy-binary in an accessible location, `$ENVOY_BINARY`.
+2. Keep all the scripts, Makefiles in a directory, `$SCRIPTS`. The scripts, Makefiles refer to the same files that are included in this benchmarking package.
+3. Keep all the Envoy configurations in a directory, `$ENVOY_CONFIG`.
+4. Select a directory in which you want to keep the result, `$RESULT`.
+
+Install the following packages (possibly running the below commands): `pexpect`
+
+1. `sudo pip install pexpect`
```
jamessynge: And I'd consider this a pre-req. Just add it above, and consider adding commands such as that you've included here for installing the other items.

Readme.md
```diff
@@ -0,0 +1,60 @@
+**Pre-requisite:**
+
+1. python2
+2. gcloud
+
+Follow these set-up before running the benchmarking script.
+
+1. Keep your envoy-binary in an accessible location, `$ENVOY_BINARY`.
+2. Keep all the scripts, Makefiles in a directory, `$SCRIPTS`. The scripts, Makefiles refer to the same files that are included in this benchmarking package.
```
jamessynge: It sounds like SCRIPTS is set to the directory that is the root of the local clone of this repo. My inclination would be to change the default, and have benchmark.py infer it from the path to itself. See these pages:

https://stackoverflow.com/questions/3718657/how-to-properly-determine-current-script-directory-in-python

http://www.faqs.org/docs/diveintopython/regression_path.html

Readme.md
```diff
@@ -0,0 +1,60 @@
...
+	  --os_img_family
+		                the OS in which you want the benchmark. (default:
+		                ubuntu-1604-lts)
+	  --os_img_project
+		                the project in which the OS can be found. (default:
+		                ubuntu-os-cloud)
+	  --project     	the project name. (default: envoy-ci)
+	  --logfile     	the local log file for this script. New log messages 
+		                are appended to this file. (default: benchmark.log)
+	 --num_retries      the number of retries for a single command. (default:
```
jamessynge: It isn't clear if retries here are for failures or for test iterations. Could you clarify?

Readme.md
sohamm17: Right, nginx is being used as the backend. However, nginx is being installed automatically by our script on the Virtual Machine in which we run our experiments.
Yes, we need to say that `lyft/envoy`'s binary is needed to provided for the experiments. I'll add that.

Readme.md
sohamm17: Yeah, sorry for being unclear. So, for some of the commands, we keep trying to execute the commands, even if they fail. One such is check whether an VM instance is running, just after it is being created. Sometimes, it may take sometime to have the instance as `Running` mode. Therefore, we need to keep checking the status of the instance. We are talking about the number of times we should retry those kinds of commands. Also, the next argument is saying that the number of seconds we should `sleep` before the next retry.

PR #7, opened 2017-06-26
Storing h2load results in a DB

sohamm17:
store_data.py can be run to store data in a GCloud SQL instance. By this Python file, an DB instance will be created on GCloud. Then, a DB and a table on the DB will be created. A h2load JSON result file will be loaded, and corresponding data will be inserted into newly created table. The table will only be created if already not exists.


store_data.py
```diff
@@ -0,0 +1,353 @@
...
+                                    "DROP DATABASE {};".format(args.database))
+    print "Database deletion is successful."
+  else:
+    print "Database deletion is skipped due to --no-delete_db."
+
+  connection.close()
+  print "DB Connection closed."
+
+if __name__ == "__main__":
+  main()
```
htuch: How does this tie in with top-level invocation of the benchmark? It would be useful to be able to say "Run h2load with revision XYZ (git hash) of Envoy 5 times and put the results in a database with distinct timestamps but related run IDs". Then, I can later go and build a query for this run ID and maybe some others, draw the graph.

htuch:
@jamessynge for additional review.

benchmark.py
```diff
@@ -8,6 +8,7 @@
 import time
 
 import pexpect
+import python_helpers as python_utils
```
htuch: I would just call this `utils` at the filename and Python level, it's a bit tautological to include Python in a Python module name.

benchmark.py
```diff
@@ -176,6 +177,35 @@ def RunBenchmark(args, logfile):
   else:
     print "Instance deletion is skipped due to --no-create_delete."
 
+  data_store_command = ("python store_data.py --ownip {}"
```
htuch: I think it's best to import `store_data` as a module and invoke the Python methods directly, rather than invoke this as an external library. This will allow a tighter integration as needed between the two.

benchmark.py
```diff
@@ -176,6 +177,35 @@ def RunBenchmark(args, logfile):
...
+    )
+  else:
+    data_store_command = "{} --no-create_db".format(data_store_command)
+
+  if args.delete_db:
+    data_store_command = ("{} --delete_db").format(
+        data_store_command
+    )
+  else:
+    data_store_command = "{} --no-delete_db".format(data_store_command)
```
htuch: Yeah, most of this can goaway if you do the above.

benchmark.py
```diff
@@ -176,6 +177,35 @@ def RunBenchmark(args, logfile):
...
+
+  print data_store_command
```
htuch: Please remove debug statements from stuff you're planning on merging.

benchmark.py
```diff
@@ -176,6 +177,35 @@ def RunBenchmark(args, logfile):
...
+  pexpect.run(data_store_command, timeout=None, logfile=logfile)
+  print "Data stored into database."
```
htuch: Can you file an issue to convert all these `print` statement to https://docs.python.org/2.7/library/logging.html? We should move away from doing `print` and do properly qualified (in terms of verbosity, warning/error/info/debug etc.) log messages.

benchmark.py
```diff
@@ -226,23 +256,46 @@ def main():
...
-                                 action="store_true",
-                                 help="if you want to create/delete new VM.")
-  create_del_parser.add_argument("--no-create_delete", dest="create_delete",
-                                 action="store_false")
+  parser.add_argument("--db_instance_name",
+                      help="the name of the gcloud instance",
+                      default="envoy-db-instance")
+  parser.add_argument("--envoy_hash",
+                      help="the hash of envoy version",
+                      default="xxxxxx")
```
htuch: There should be no default, it should be an error not to specify.

benchmark.py
```diff
@@ -226,23 +256,46 @@ def main():
...
+  parser.add_argument("--runid",
+                      help="the run id of this benchmark",
+                      default="0")
+  parser.add_argument("--ownip", help=("the machine's IP from where"
+                                       " the script is being run"),
+                      default="127.0.0.1")
```
htuch: This IP based authentication seems weird. Is there no better way to do credential based authentication to Cloud SQL? If not, you should be able to automatically infer this `--ownip` in any case.

benchmark.py
```diff
@@ -226,23 +256,46 @@ def main():
...
+  parser.add_argument("--database", help="name of the database",
+                      default="envoy_stat_db")
+
+  python_utils.CreateMutuallyExclusiveArgument(parser, "create_delete",
```
htuch: BTW, I would rename this to `CreateBooleanArgument`.

benchmark.py
```diff
@@ -226,23 +256,46 @@ def main():
...
+
+  python_utils.CreateMutuallyExclusiveArgument(parser, "create_db",
+                                               ("turn on if you want"
+                                                " to create the DB"))
+  parser.set_defaults(create_db=True)
+
+  python_utils.CreateMutuallyExclusiveArgument(parser, "delete_db",
+                                               ("turn on if you want"
+                                                " to delete the DB"))
+  parser.set_defaults(delete_db=True)
```
htuch: Why can't this be integrated into `CreateBoolArgument`?

store_data.py
```diff
@@ -0,0 +1,355 @@
...
+
+def GetByteData(data):
+  """The function accepts a dictionary and returns Bytes value inside it.
+
+  Based on the unit (B, MB, KB), the function converts to Bytes value. It also
+  works for B/s, MB/s, etc.
+  Args:
+    data: data is a dictionary of the format: {data: <val>, unit: <unit>}.
+  Returns:
+    A float value of the Bytes unit.
```
htuch: integer

store_data.py
```diff
@@ -0,0 +1,355 @@
...
+
+  if data["unit"].startswith("B"):
+    return int(data["data"])
+  elif data["unit"].startswith("K"):
+    return int(float(data["data"]) * 1024)
+  elif data["unit"].startswith("M"):
+    return int(float(data["data"]) * 1024 * 1024)
+
+
+def JSONToInsertCommand(table, **kwargs):
```
htuch: This should have a more descriptive name - it's not working with JSON at all (the caller is). E.g. `InsertCommandFromArgs`.

store_data.py
```diff
@@ -0,0 +1,355 @@
...
+  Returns:
+    A float value of the Bytes unit.
+  """
+  assert (data["unit"].startswith("B") or data["unit"].startswith("K") or
+          data["unit"].startswith("M"))
+
+  if data["unit"].startswith("B"):
+    return int(data["data"])
+  elif data["unit"].startswith("K"):
+    return int(float(data["data"]) * 1024)
```
htuch: Why convert to float and then to int? I think you can just do `int(data["data"])`.

benchmark.py
sohamm17: I figured out that it could be. Done.

store_data.py
sohamm17: To prevent some data-loss, I think. I mean the value can 0.5 KB. So, 0.5*1024=512 Byte would be correct.

benchmark.py
```diff
@@ -176,10 +183,49 @@ def RunBenchmark(args, logfile):
...
+    data_store_command = ("{} --create_db --database {}").format(
+        data_store_command, args.database)
+  else:
+    data_store_command = "{} --no-create_db".format(data_store_command)
+
+  if args.delete_db:
+    data_store_command = ("{} --delete_db").format(
+        data_store_command)
+  else:
+    data_store_command = "{} --no-delete_db".format(data_store_command)
```
htuch: There's a lot of boiler plate for the yes/no case here, can you refactor into a general method for inserting "--no-" in front as needed?

benchmark.py
```diff
@@ -226,22 +268,52 @@ def main():
...
-                                 action="store_false")
-  parser.set_defaults(setup=True)
+  parser.add_argument("--db_instance_name",
+                      help="the name of the gcloud instance",
+                      default="envoy-db-instance")
+  parser.add_argument("--tier", help="the tier of GCloud SQL service",
+                      default="db-n1-standard-2")
+  parser.add_argument("--db_username", help="username on the DB",
+                      default="root")
+  parser.add_argument("--db_password",
```
htuch: Passwords on command-line are dangerous (since anyone can view them via ps). Are there alternatives? An environment variable might make sense, not sure.

database_helpers.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def ExecuteAndReturnResult(connection, statement):
+  """This function will execute a SQL statement and return the SQL output.
+
+  Args:
+    connection: connection to the DB instance
+    statement: statement to be executed.
+  Returns:
+    It returns the SQl output after executing the `statement`.
```
htuch: Nit: s/SQl/SQL

store_data.py
```diff
@@ -0,0 +1,341 @@
...
+
+def CreateTable(connection, table_name):
+  """This function creates the envoy_stat table on the give `connection`.
+
+  Args:
+    connection: the connection which will refer to a database in which the table
+    will be created.
+    table_name: the name of the table to be created.
+  Returns/Yields:
+    It creates a table on the `connection`.
```
htuch: This isn't a Returns/Yields comment, it belongs above. This function doesn't return or yield any exception.

store_data.py
```diff
@@ -0,0 +1,341 @@
...
+  and the table-structure.
+  Args:
+    connection: the connection with the database
+    table: name of the table in which `json_file` will be entered into
+    json_file: an opened filestream in reading mode. The file needs to be in
+    the correct JSON format, generated by the benchmarking script.
+    runid: runid of the result
+    envoy_hash: the hash of the envoy version used for this result
+  Returns/Yields:
+    It inserts the json_file into the `table`, by iterating the file.
```
htuch: Same as above on returns/yields. Can you fix this everywhere? Thanks.

store_data.py
```diff
@@ -0,0 +1,341 @@
...
+    print "Google Cloud SQL Instance {} is created.".format(
+        args.db_instance_name)
+  else:
+    print "DB Instance creation is skipped due to --no-create_instance."
+
+  sh_utils.RunGCloudService(["users", "set-password", args.username, "%",
+                             "--instance", args.db_instance_name, "--password",
+                             args.password], project=args.project,
+                            service="sql", logfile=logfile)
+  print "DB Usernames and passwords are set."
```
htuch: Wait a sec, can't we dynamically generate the password in the script here and then pass to gcloud and to the MySQLdb connection? That way we never need to put it anywhere persistent like the environment or command-line. That seems the safest.

htuch: Make it large! I would make it at least 20 chars.. :)

benchmark.py
```diff
@@ -73,6 +83,32 @@ def TryFunctionWithTimeout(func, error_handler, num_tries,
   raise BenchmarkError("All tries failed.")
 
 
+def AppendArgument(main_command, boolean_var, arg_name,
```
htuch: I would change this to `GetBooleanFormattedArgument` and drop the append part (just return a string with the arg, let the caller `+=` it). That would be more functional and avoid baking in policy in the function.

benchmark.py
```diff
@@ -176,10 +209,41 @@ def RunBenchmark(args, logfile):
   else:
     print "Instance deletion is skipped due to --no-create_delete."
 
+  ownip = StringIO.StringIO()
+  sh_utils.RunSSHCommand(args.username, args.vm_name,
```
htuch: I feel it would probably make sense to have a persistent SSH connection (under pexpect) for this sort of thing. That way, you don't need to keep re-SSHing to the VM each time you need to do something. That would probably be a noticeable performance improvement. Please file an issue for this.

benchmark.py
```diff
@@ -226,22 +286,49 @@ def main():
...
+                              ("turn on if you want to create"
+                               " a Google Cloud SQL instance"),
+                              create_db_instance=True)
+
+  utils.CreateBooleanArgument(parser, "create_db",
+                              ("turn on if you want"
+                               " to create the DB"),
+                              create_db=True)
+
+  utils.CreateBooleanArgument(parser, "delete_db",
```
htuch: I would remove this from the main benchmark utility. There should maybe be an independent utility for this, but it doesn't make sense in the context of the benchmark.

store_data.py
```diff
@@ -0,0 +1,337 @@
+"""This file creates is for the DB-related work on the statistics."""
```
htuch: Nit: Please reword the start of this sentence.

utils.py
```diff
@@ -0,0 +1,42 @@
...
+import string
+
+def CreateBooleanArgument(parser, argument_name, help_string,
+                          **default_condition):
+  """The function creates a mututally exclusive argument on parser.
+
+  It just creates an argument against the `argument_name` on parser with a
+  required help_string. The given `argument_name` is set to True and the
+  --no-`argument_name` is set to False. It doesn't set the default value. The
+  caller needs to set the default against `argument_name`.
```
htuch: Is the last sentence here true now that you do `set_defaults` below?

utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def GetRandomPassword():
+  """This function generates a random 20-length password.
+
+  Returns:
+    Returns a random 20-length password consisting of ASCII characters.
+  """
+  length = 20
+  random.seed = (os.urandom(1024))
```
htuch: Nit: no need for parentheses here.

benchmark.py
sohamm17: Well, the `utils.py` is kind of a generic utility for Python. At least, that's why I created it.

utils.py
sohamm17: No, my bad. Sorry!

PR #10, opened 2017-06-30
Generating graph with the data from the database

sohamm17:
Graph, related to Run ID or Envoy Hash, can be generated by running the show_graph.py Python script. This is a separate script because multiple runids and Envoy hashes can be provided.


database_helpers.py
```diff
@@ -52,3 +52,17 @@ def GetFieldFromTable(connection, table_name, field="*", cond=None):
...
+
+  Args:
+    select_based_list: the select based list.
+  Returns:
+    A Python list
+  """
+  converted_list = list()
+  for data in select_based_list:
+    converted_list.append(data[0])
+  return converted_list
```
htuch: This could be written as something like `return [d[0] for d in select_based_list]`, or even as `return map(operator.itemgetter(0), selected_based_list)`

show_graph.py
```diff
@@ -0,0 +1,148 @@
+"""This file creates a graph from the metric Gcloud SQL database."""
```
htuch: Metric as opposed to imperial? ;)

show_graph.py
```diff
@@ -0,0 +1,148 @@
...
+
+import argparse
+
+import database_helpers as db_utils
+import matplotlib.pyplot as plt
```
htuch: Do you want to add matplotlib to the setup instructions?

show_graph.py
```diff
@@ -0,0 +1,148 @@
...
+    direct_lists.append(direct_list)
+
+    condition = ("where {}=\"{}\" and"
+                 " category=\"envoy\"").format(x_axis_field, x)
+    envoy_list = db_utils.SingleColumnToList(db_utils.GetFieldFromTable(
+        connection, table_name, field=y_axis_field, cond=condition))
+    if not envoy_list:
+      print "{} {} is not found in table for Envoy results.".format(
+          x_axis_field, x)
+      envoy_list = [0]
```
htuch: The code above is duplicated, it would make sense to refactor into an internal function (not at top level).

show_graph.py
```diff
@@ -0,0 +1,148 @@
...
+  direct_std = list()
+  envoy_means = list()
+  envoy_std = list()
+
+  for single_set in direct_lists:
+    direct_means.append(np.mean(single_set))
+    direct_std.append(np.std(single_set))
+  for single_set in envoy_lists:
+    envoy_means.append(np.mean(single_set))
+    envoy_std.append(np.std(single_set))
```
htuch: Same here. Also, list comprehensions or maps are probably more Pythonic than looping.

show_graph.py
```diff
@@ -0,0 +1,148 @@
...
+  rects2 = ax.bar(ind + width, envoy_means, width, color="y", yerr=envoy_std)
+
+  ax.set_ylabel(y_axis_field)
+  ax.set_xlabel(x_axis_field)
+  ax.set_xticks(ind + width)
+  ax.set_xticklabels(x_axis_values)
+  ax.legend((rects1[0], rects2[0]), ("Direct", "Envoy"))
+  AutoLabel(rects1, ax)
+  AutoLabel(rects2, ax)
+  fig.savefig("{} {}.pdf".format(
```
htuch: FYI, I think using PNG or some other image format is more helpful for sharing in general at Google.

show_graph.py
```diff
@@ -0,0 +1,148 @@
...
+                      nargs="+")
+  parser.add_argument("--fieldname",
+                      help="the name of the field for which you want the graph",
+                      default="total_req_per_sec")
+
+  args = parser.parse_args()
+  logfile = open(args.logfile, "ab")
+  hostname = db_utils.GetInstanceIP(args.db_instance_name, args.project)
+
+  password = utils.GetRandomPassword()
```
htuch: All this database setup stuff should be refactored, it looks weird to be setting a password to generate a graph.

htuch:
@jamessynge for a review pass.

show_graph.py
sohamm17: This is actually authorization to the database. As this graph generation script can be run from any other machine, not only the benchmarking machine, I needed to authorize the DB again for the new machine. Instead of saving the previous random password from the previous scripts, I thought of setting a new random password as safer.

htuch: Yep, makes sense. I'm just suggesting you put this authorization setup code in a common place so it doesn't have to be duplicated across different modules.

Readme.md
```diff
@@ -58,8 +58,37 @@ The above command will create a VM in the `us-east1-b` zone (default) with the n
...
+	  --setup           if you want to run setup on benchmarking VM. (default:
+		                True)
 	  --no-setup
-	                        
\ No newline at end of file
+	  --create_db_instance  
+	  					turn on if you want to create a Google Cloud SQL
+		                instance (default: True)
+	  --no-create_db_instance
+	  --create_db       turn on if you want to create the DB (default: True)
```
htuch: The graphing code should never need to create a DB - if these parameters are needed, it should be possible to synthesize them internally in the graph script.

show_graph.py
```diff
@@ -21,44 +18,62 @@ def DrawBarGraph(connection, table_name, y_axis_field, x_axis_values,
...
-    if not envoy_list:
-      print "{} {} is not found in table for Envoy results.".format(
-          x_axis_field, x)
-      envoy_list = [0]
-
-    envoy_lists.append(envoy_list)
+
+  def GetListsFromDB(x_axis_values, x_axis_field, connection,
+                     table_name, y_axis_field, category):
+    """This function returns lists of values of a field from the DB.
```
htuch: No need to change anything, but for future reference, there's no need to write details function-level comments for internal functions unless it's really unclear what's going on - you can just add a few line # comment above explaining what it does. The full arg/returns docs are for module-level functions that are likely to be exported out and consumed by others.

show_graph.py
```diff
@@ -21,44 +18,62 @@ def DrawBarGraph(connection, table_name, y_axis_field, x_axis_values,
...
-  for single_set in envoy_lists:
-    envoy_means.append(np.mean(single_set))
-    envoy_std.append(np.std(single_set))
+
+  def GetMeansAndStdsFromList(lists):
+    """This function returns the means and standard deviation of lists.
+
+    Args:
+      lists: A list of lists. Each list inside the top-level list consists a
+      bunch of numbers
```
htuch: ".. consists of a sample for a given variable that summary stats will be computed on."

show_graph.py
```diff
@@ -21,44 +18,62 @@ def DrawBarGraph(connection, table_name, y_axis_field, x_axis_values,
...
+    Returns:
+      A pair of list containing means and standard deviations.
+    """
+    means = [np.mean(single_list) for single_list in lists]
+    stds = [np.std(single_list) for single_list in lists]
+    return (means, stds)
```
htuch: Nit: `return means, stds`

Readme.md
sohamm17: Yeah, this parameter is for the `benchmark.py` script, not the `show_graph.py` script. I forgot to add those in the earlier PRs.

PR #12, opened 2017-07-10
Minor changes for setup

sohamm17:
Also, incorporated James' suggestions


sohamm17:
@htuch

htuch:
Can you merge to resolve conflicts @sohamm17?

sohamm17:
@htuch I have made all the changes needed for this PR.

utils.py
```diff
@@ -40,3 +40,14 @@ def GetRandomPassword():
...
+
+def ParseCommaSeparatedTuple(comma_sep_string):
+  """This function parses out a comma-separated string to two separate values.
+
+  Args:
+    comma_sep_string: the comma-separated string.
+  Returns:
+    A tuple of two separated values
+  """
+  return comma_sep_string.split(",")
```
htuch: I would just inline this, it's quite Pythonic to `split(',')` and more concise.

benchmark.py
```diff
@@ -300,6 +348,30 @@ def main():
...
+                           "run, separated by a comma.",
+                      default="0,10")
+  parser.add_argument("--envoy_cores",
+                      help="the start and end core numbers for"
+                           " Envoy to run, separated by a comma.",
+                      default="11,18")
+  parser.add_argument("--h2load_cores",
+                      help="the start and end core numbers for "
+                           "h2load to run, separated by a comma.",
+                      default="19,19")
```
htuch: The default seems somewhat imbalanced. Having a single core for client load generation seems likely to bottleneck at the client. Can we adjust these defaults based on your current findings?

Readme.md
```diff
@@ -1,9 +1,12 @@
...
-1. python2
-2. gcloud
+1. python2.7 Installation: `sudo apt-get install python2.7`
+2. gcloud Installation: Follow instructions here: https://cloud.google.com/sdk/docs/quickstart-debian-ubuntu
+3. Install the following packages:
+`sudo pip install pexpect`
+`sudo apt-get install python-matplotlib`
 
-Follow these set-up before running the benchmarking script.
+Edit and set the values of these variables (e.g. `$ENVOY_BINARY`) to the appropriate values:
```
sohamm17: We wanted to say that modify the environment variables.

benchmark.py
```diff
@@ -67,8 +67,8 @@ def TryFunctionWithTimeout(func, error_handler, num_tries,
   count = num_tries
   while count > 0:
     try:
-      ret_val = func(*args, **kwargs)
```
sohamm17: Yes, the count was otherwise, not updated, before the function call.

benchmark.py
sohamm17: Actually, I tuned this so that h2load is given only 1 core, as our primary experimentation shows more cores to h2load increases variance. nginx is given 10 cores for 10 worker processes, by default. Rest of the cores are given to Envoy.

benchmark.py
```diff
@@ -233,6 +266,19 @@ def RunBenchmark(args, logfile):
                          logfile=logfile, zone=args.zone, project=args.project)
   print "Data stored into database."
 
+  if args.create_delete:
+    print "Deleting instance. Wait..."
+    # pexpect.run does not take argument as arrays
+    pexpect.run(("gcloud compute --project {}"
+                 " instances delete {} --zone {}").format(
+                     args.project, args.vm_name, args.zone),
+                events={"Do you want to continue (Y/n)?": "Y\n"},
```
htuch: BTW, you shouldn't need to interactively do this Y/N business with gcloud, it has a `--quiet` setting to allow batch execution. See `--quiet` in https://cloud.google.com/sdk/gcloud/reference/. Can you file an issue or a PR to clean this up?

benchmark.py
htuch: OK, we can change this as we go. I think we'll need more than one core now that we have decent looking results from Envoy.

PR #14, opened 2017-07-26
Support for option of including ssl; Incorporating mpstat logging; New h2load setup

sohamm17:


sohamm17:
Not generated any graph yet. I'll do that.

sohamm17:
Forgot to change the URL provided to h2load. @htuch

PR #15, opened 2017-07-26
Added support for generating graphs based on starting time

sohamm17:


show_graph.py
```diff
@@ -137,6 +204,12 @@ def main():
   parser.add_argument("--fieldname",
                       help="the name of the field for which you want the graph",
                       default="total_req_per_sec")
+  parser.add_argument("--arrangement", help=("the type of arrangement in"
+                                             " this experiment."),
```
htuch: I think you need to better document in various places (and here) what an arrangement is.

show_graph.py
```diff
@@ -108,6 +111,70 @@ def AutoLabel(rects, ax):
...
+    direct_means = [v[0] for v in direct_list]
+    direct_std = [v[1] for v in direct_list]
+    direct_times = [v[2].time().strftime("%H:%M") if not i % 2 else ""
+                    for i, v in enumerate(direct_list)]
+  else:
+    return
+
+  if envoy_list:
+    envoy_means = [v[0] for v in envoy_list]
+    envoy_std = [v[1] for v in direct_list]
```
htuch: Here (and above), you can write this more cleanly. Something like: `envoy_means, envoy_std = zip(*v)`

show_graph.py
```diff
@@ -108,6 +111,70 @@ def AutoLabel(rects, ax):
...
+    # time is not needed again but if needed, it can be taken from here
+    # envoy_times = [v[2] for v in envoy_list]
+  else:
```
htuch: Are we silently failing here (same above)?

sohamm17: Yeah, that was not elegant. Sorry.

show_graph.py
sohamm17: That was neat. Thanks.

PR #17, opened 2017-08-16
Enable HTTP 1.1

sohamm17:


distribute_proc.py
```diff
@@ -198,6 +198,9 @@ def RunAndParseH2Load(h2load_command, h2load_timeout=None, logfile=None):
       "sd%": float(rps_sd_per)
   }
 
+  while child.isalive():
+    print "h2load is still alive, after parsing is complete."
```
htuch: Should you have a sleep in there?

PR #18, opened 2017-08-16
Circuit breakers in Simple Loopback

sohamm17:


simple-loopback.json
```diff
@@ -54,6 +54,13 @@
       {
         "name": "local_service",
         "features": "http2",
+        "circuit_breakers": {
+          "default": {
+            "max_connections": 98304,
```
htuch: You should just max these out. This won't cause Envoy to create all these connections.

PR #22, opened 2018-10-08
perf: Move old perf files to a subdir and add a new README.

jmarantz:
Moved all the old perf files to a subdir. These old files (a) didn't work as advertised in the README (b) appeared to be addressing one angle of the perf challenge and I am at this point interested in a different one.

It's possible we may pick up the effort started with these files again, and that can be done in the context of the subdirectory.


dnoe:
For some reason GitHub shows this as just deletions, no move. Was that the intent? Is GH being screwy or did something else get messed up?

jmarantz:
Yeah I had forgotten to 'git add' the new subdir; fixed now.

PR #24, opened 2018-10-08
Add files needed for siege-based tests.

jmarantz:
Adds initial version of siege-based tests, using Envoy as an origin and a proxy. There's still a lot to do, but so far it:
 - has one canned configuration
 - expects two binaries: a clean and an experimental one
 - runs a bunch of sieges interleaved between the two binaries
 - prints out some basic stats in a table showing which binary is faster and smaller, including stddev
    across multiple runs.

Still to do:
 - better UI for specifying the binaries
 - lots more configs to wake up different aspects of scalability (ssl, large requests, large numbers of
   configs, stats sinks, etc)
 - maybe annotate the table of results with some more stat values scraped from admin.

Example output:
```
               Clean          Std Dev        Experimental   Std Dev        Improvement    
               -----          -------        ------------   -------        -----------    
Trans Rate     32145.16       3314.479       31738.48       2655.239       1.265%         
Throughput     105.44         10.872         104.1          8.707          1.265%         
Failed         0.0            0.0            0.0            0.0            0              
Stats Mem      4209331.2      15651.771      4203342.4      15465.695      0.142%         
VSZ            343390.4       2.191          343468.8       1.789          -0.023%        
RSS            27199.2        326.465        27544.0        285.04         -1.268%        
```

Signed-off-by: Joshua Marantz <jmarantz@google.com>


jmarantz:
closing this as DCO workarounds seemed to not work...and shouldn't be needed going forward.

PR #25, opened 2018-10-08
siege: Initial checkin of siege scripts to test envoy perf and memory use.

jmarantz:
Adds initial version of siege-based tests, using Envoy as an origin and a proxy. There's still a lot to do, but so far it:
 - has one canned configuration
 - expects two binaries: a clean and an experimental one
 - runs a bunch of sieges interleaved between the two binaries
 - prints out some basic stats in a table showing which binary is faster and smaller, including stddev
    across multiple runs.

Still to do:
 - better UI for specifying the binaries
 - lots more configs to wake up different aspects of scalability (ssl, large requests, large numbers of
   configs, stats sinks, etc)
 - maybe annotate the table of results with some more stat values scraped from admin.

Example output:
```
            Clean      Std Dev    Experimental  Std Dev    Improvement
            -----      -------    ------------  -------    -----------
Trans Rate  37313.43   1937.149   33783.79      1937.626   9.459%
Throughput  122.39     6.352      110.81        6.357      9.462%
Failed      0.0        0.0        0.0           0.0        0
EnvoyMem    4228976.0  16643.443  4187808.0     15852.695  0.973%
VSZ         343388.0   2.191      343472.0      2.191      -0.024%
RSS         51604.0    834.403    50684.0       2546.639   1.783%
```


jmarantz:
@mattklein123 @htuch @alyssawilk would love feedback on this. I'll add some more comments in the meantime.

siege/front-proxy.yaml
```diff
@@ -0,0 +1,69 @@
...
+              routes:
+              - match:
+                  prefix: "/"
+                direct_response:
+                  status: 200
+                  body:
+                    filename: "./lorem_ipsum.txt"
+          http_filters:
+          - name: envoy.router
+            config: {}
```
htuch: Do you want https://github.com/envoyproxy/envoy-perf/blob/master/cloudperf/simple-loopback.json#L38 for these routers configs? Even when comparing relative performance, it's best not to skew the absolute numbers with things that are known not to be set when running high-performance benchmarks.

htuch: Same also for https://github.com/envoyproxy/envoy-perf/blob/master/cloudperf/simple-loopback.json#L42

jmarantz:
ptal...some unresolved issues exist; looking for more info on some config tweaks, and would like to rewrite driver script in Python as a follow-up.

siege/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. Run performance benchmarks on optimized builds: `bazel build -c opt source/exe:envoy-static`.
+2. Don't let any other software run on the machine at the same time.
+3. Make sure the configuration you are testing actually hits the code you are editing.
+4. If you are testing code changes, interleave runs between an envoy-static built from master
+   and one built from your branch.
+
+##About Siege:
+
+Siege is a github project that efficiently pummels an HTTP server with requests and measures
```
alyssawilk: Is Siege HTTTP/1 only?  I notice it's 1.1 in your config below and if so I think it's worth call out specifically here.

jmarantz:
Thanks @alyssawilk -- HTTP/1.x only -- I put HTTP 1.1 in the README. H2 is not in the codebase AFAICT and so this can't be the final answer for all aspects of perf for us. There are likely other limitations as well which we'll explore going forward.

jmarantz:
Variance is higher than what I would expect, but not terrible. As an example, in the most recent run I did with a clean Envoy, the results are:

37313.43
33670.04
36231.88
38167.94
38461.54

I think one of the challenges going forward will be to reduce this. The variance is higher than I've seen for microbenchmarks like https://github.com/envoyproxy/envoy/blob/master/test/common/common/utility_speed_test.cc

siege/siege.sh
```diff
@@ -0,0 +1,134 @@
...
+
+# We use the admin port both for scraping memory and for quitting at
+# the end of each run.
+admin="http://127.0.0.1:$admin_port"
+
+# Blocks the script until a server started in the background is ready
+# to respond to the URL passed as an arg.
+function wait_for_url() {
+  local url="$1"
+  (set +e; set +x; until curl "$url"; do sleep 0.1; done) &>> "$log"
```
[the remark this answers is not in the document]
htuch: This is OK, it's definitely doable without much hassle in Python, feel free to chat to me in IRL tomorrow how to do it (some combination of subprocess.check, Popen etc.). Soham's scripts can provide some reference here as well.

siege/siege_result_analysis.py
```diff
@@ -0,0 +1,101 @@
...
+    sys.exit(1)
+
+  # Save the performance CSV data from the clean and experimental runs, using
+  # an array of maps.
+  clean = parseCsv(argv[1])
+  experimental = parseCsv(argv[3])
+
+  # Set up a matrix for printing the analyzed perf results.
+  matrix = [["", "Clean", "Std Dev", "Experimental", "Std Dev", "Improvement"],
+            ["", "-----", "-------", "------------", "-------", "-----------"]]
```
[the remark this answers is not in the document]
htuch: Yeah, as long as we have machine generated output as well I think that's good.

What's the opposition to doing a venv setup and using one of the existing CSV table formatter in Python? We already have a bunch of examples of venv in the main repo. IMHO, this is basically Python best practice, since you then make available a ton of others libraries at low cost and avoid future developers of this script from having to manually craft things that are already done elsewhere. The benefits of being standalone are relatively low, since venv/pip are totally standard.

siege/front-proxy.yaml
htuch: Disable the request UUID generation and dynamic stats is considered best practice when benchmarking Envoy's data plane. Of course, one day maybe we want to benchmarking them as well, but I would make that a TODO for when we have templated configuration generation, which is definitely where we want to take this.

siege/siege_result_analysis.py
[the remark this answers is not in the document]
jmarantz: No opposition; just trying to be expedient. The 'column' command was revealed to me in the stackoverflow link you shared and was pretty simple.

However when I python-ize the main script I'll switch to using native Python rather than 'column'.

README.md
```diff
@@ -1,6 +1,6 @@
 # Envoy performance tools collection
 
-**Performance benchmarking Options**
+##Performance benchmarking Options:
```
htuch: This looks wrong in Markdown preview. I think you want a space after the `##` (And no : at the end is needed)

siege/siege_result_analysis.py
```diff
@@ -0,0 +1,108 @@
...
+           'experimental_mem_csv aggregate_csv') % argv[0])
+    sys.exit(1)
+
+  # Save the performance CSV data from the clean and experimental runs, using
+  # an array of maps.
+  clean = parseCsv(argv[1])
+  experimental = parseCsv(argv[3])
+
+  # Write aggregated stats to aggregate.csv
+  aggregate = open(argv[5], "w")
```
htuch: Best to use context managers, e.g. `with open(...) as f:` in Python for file I/O.

siege/front-proxy.json
```diff
@@ -0,0 +1,136 @@
...
+  "static_resources": {
+    "listeners": [
+      {
+        "address": {
+          "socket_address": {
+            "address": "0.0.0.0",
+            "port_value": 10000
+          }
+        },
+        "filter_chains": [
```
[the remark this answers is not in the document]
jmarantz: I really don't love either, but the Envoy doc is in JSON and I can mentally map JSON to the protobuf syntax more easily.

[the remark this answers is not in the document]
htuch: Which docs? We've been migrating all the docs to YAML..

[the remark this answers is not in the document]
jmarantz: https://github.com/envoyproxy/envoy-perf/blob/master/cloudperf/simple-loopback.json#L38 which you pointed me to above.

Also https://www.envoyproxy.io/docs/envoy/latest/api-v1/network_filters/http_conn_man which I found with a google search [envoy proxy generate_request_id] -- also inspired by what you referenced above.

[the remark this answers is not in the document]
htuch: You pointed at two v1 JSON API resources, which have now had support dropped officially :)

jmarantz:
@ambuc FYI

siege/front-proxy.yaml
```diff
@@ -0,0 +1,75 @@
...
+          stat_prefix: ingress_http
+          route_config:
+            name: local_route
+            virtual_hosts:
+            - name: service
+              domains:
+              - "*"
+              routes:
+              - match:
+                  prefix: "/"
```
htuch: Nit: YAML doesn't need `".."`

siege/front-proxy.yaml
```diff
@@ -0,0 +1,75 @@
...
+                direct_response:
+                  status: 200
+                  body:
+                    filename: "./lorem_ipsum.txt"
+          http_filters:
+          - name: envoy.router
+            config: {
```
htuch: Nit: braces here are redundant in YAML

README.md
jmarantz: done

jmarantz:
@htuch ptal

PR #27, opened 2018-10-14
siege: Rewrite siege.sh in python

jmarantz:
This is an attempt at a direct rewrite of siege.sh in Python. More could be done to make it idiomatic Python, comments welcome @htuch @zuercher


siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+  # the runs we are about to do.
+  for fname in csv_files + [log]:
+    if os.path.exists(fname):
+      os.remove(fname)
+
+  with open(log, "a") as logfile:
+    # Echoes a command to the logfile, and then returns it. This is
+    # intended to wrap the command argment to subprocess.Popen and
+    # subprocess.call so we can see in the log how they were run.
+    def echo(command):
```
htuch: This seems to be reinventing `strace`. Why not just use an `strace` filtered down to `exec` calls to obtain the same?

siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+# JSON object.
+def loadJson(url):
+  handle = urllib.request.urlopen(url)
+  data = handle.read().decode("utf-8")
+  handle.close()
+  return json.loads(data)
+
+# Runs a shell command, returning its stdout as a string.
+def runCapturingStdout(command):
+  proc = subprocess.run(command, shell=True, stdout=subprocess.PIPE)
```
htuch: Why not just `subprocess.check_output`?

siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+  # outside the try/except block and let the exception propagate.
+  readUrl()
+
+# Reads text output from a URL and decodes it as JSON, returning the
+# JSON object.
+def loadJson(url):
+  handle = urllib.request.urlopen(url)
+  data = handle.read().decode("utf-8")
+  handle.close()
+  return json.loads(data)
```
htuch: Can this be done as a context managed block?

siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+    # Loops 5x interleaving runs with the clean and experimental versions
+    # of Envoy, collecting the results in separate CSV files.
+    progress("Logging 10 runs to %s " % log)
+    for run in range(0, 4):
+      runEnvoyAndSiege(clean_envoy, clean_csv_files)
+      progress("...%d" % (2 * run + 1))
+      runEnvoyAndSiege(experimental_envoy, experimental_csv_files)
+      progress("...%d" % (2 * run + 2))
+
+    print("\n")
```
htuch: Not a super huge fan of verbose plain text based output unless guarded by something like `-d` for debug. Ideally these are tools we are building for automation, so we should think in terms of what would be machine consumed.

siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+  #
+  # TODO(jmarantz): it would be nice to make the number of reps configurable,
+  # e.g. with a command-line option.
+  reps = 2000
+  siege_args = ["--reps=%d" % reps, "--rc=" + siege_dir + "/siege.conf", 
+                proxy_url]
+
+  # Clean up any old log files, so the CSVs just show the results from
+  # the runs we are about to do.
+  for fname in csv_files + [log]:
```
htuch: Do we really need this step if doing `"w"` overwriting below?

siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+  # Connect to the script directory so local references to the yaml configs and
+  # lorem_ipsum.txt work.
+  siege_dir = os.path.abspath(os.path.dirname(argv[0]))
+  os.chdir(siege_dir)
+
+  # Derive some temp filenames from $outdir.
+  log = outdir + "/siege.log"
+  clean_perf_csv = outdir + "/clean.perf.csv"
+  clean_mem_csv = outdir + "/clean.mem.csv"
+  clean_csv_files = [clean_perf_csv, clean_mem_csv]
```
htuch: Use `os.path.join` for all of these.

siege/siege.py
```diff
@@ -0,0 +1,194 @@
...
+admin = "http://127.0.0.1:%d" % admin_port
+
+def main(argv):
+  if len(argv) != 4:
+    print("Usage: %s clean_envoy_binary experimental_envoy_binary output_dir"
+          % argv[0])
+    sys.exit(1)
+
+  # Capture arguments as locals for readability.
+  clean_envoy = argv[1]
```
htuch: I think that with the number of possible configuration options, e.g. ports, debug/quiet modes as well as positional args, that `argparse` would make sense (https://docs.python.org/3/library/argparse.html).

siege/siege.py
jmarantz: Not needed for the memory csv files because we are writing those. But if we don't clear the perf csvs then they'll be aggregated from previous runs by siege. I'll clarify in the comment.

siege/siege.py
htuch: I mean use strace externally. I.e. since you only need this for debug, just run `strace -e trace=execve -f <your program` to see the sequence of exec calls. This avoids reimplementing strace verbosely.

jmarantz: TBH that seems less portable compared to just echoing the arg array to the logfile. And also i think it means I don't get easy access to the envoy pid which I use below.

Do you feel strongly about this?

jmarantz: TBH that seems less portable compared to just echoing the arg array to the logfile. And also i think it means I don't get easy access to the envoy pid which I use below.

Do you feel strongly about this?

htuch: Yeah, because it uglifies Python code with explicit tracing everywhere :) It's less robust (you need to manually annotate) and more verbose. `strace` can give you the Envoy PID

PR #38, opened 2019-01-25
Nighthawk: A benchmarking tool built upon Envoy's libraries.

oschaaf:
This contains the inial code-drop, and addresses most of the comments from the first round of review.


oschaaf:
Close and reopen for triggering CI

htuch:
@oschaaf do you want to try setting resource limits in the Bazel build? See https://github.com/envoyproxy/envoy/pull/5698/files and the lines with `host_jvm_args` and `local_resources` for an example. This is an active fire we are trying to fight in Envoy-land.

oschaaf:
@htuch integrated and tested clang-tidy here, test was executed and passed.
On a side-note, the test took pretty long, and that is because it looks like the CI throttling actually works in that test.

oschaaf:
@htuch I thought it would be nice to add a valgrind in CI but I see this odd failure over at [1]:
```
[ RUN      ] IpVersions/BenchmarkClientTest.BasicTestH1WithRequestQueue/IPv4
vex amd64->IR: unhandled instruction bytes: 0x48 0xF 0xC7 0xF1 0x73 0x14 0x48 0x89
```

Some searching for that makes me think we're running into https://bugs.kde.org/show_bug.cgi?id=401846...
What do you think, should testing with `valgrind` be pursued, or should the valgrind test be backed out from CI for now? 

[1] https://circleci.com/gh/envoyproxy/envoy-perf/194

oschaaf:
Closing this PR, as we merged everything in pieces via other PR's.

PR #44, opened 2019-02-08
Run clang-tidy

oschaaf:


tools/gen_compilation_database.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python
```
htuch: Ideally we move the clang-tidy support and other shared code that both Envoy main repo ad envoy-perf depend upon to a submodule. @oschaaf can you file an issue for this on the main Envoy repo? @lizan does this sound like a good idea?

oschaaf: Issue was filed, marking this as resolved

PR #45, opened 2019-02-08
Nighthawk: options, rate_limiter, and utility

oschaaf:

### DIFF
--- a/nighthawk/source/client/options.proto
+++ b/nighthawk/source/client/options.proto
@@ -4,6 +4,8 @@ package nighthawk.client;
 
 import "google/protobuf/duration.proto";
 
+// TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
+// can arrive via CLI is just a concrete detail. Change this to reflect that.
 message CommandLineOptions {
   // See :option:`--rps` for details.
   uint64 requests_per_second = 1;
@@ -19,6 +21,6 @@ message CommandLineOptions {
   string concurrency = 6;
   // See :option:`--verbosity` for details.
   string verbosity = 7;
-
+  // See :option:`--uri` for details.
   string uri = 8;
 }
--- a/nighthawk/source/client/options.proto
+++ b/nighthawk/source/client/options.proto
@@ -1,0 +1,24 @@
+syntax = "proto3";
+
+package nighthawk.client;
+
+import "google/protobuf/duration.proto";
+
+message CommandLineOptions {
+  // See :option:`--rps` for details.
+  uint64 requests_per_second = 1;
+  // See :option:`--connections` for details.
+  uint64 connections = 2;
+  // See :option:`--duration` for details.
+  google.protobuf.Duration duration = 3;
+  // See :option:`--timeout` for details.
+  google.protobuf.Duration timeout = 4;
+  // See :option:`--h2` for details.
+  bool h2 = 5;
+  // See :option:`--concurrency` for details.
+  string concurrency = 6;
+  // See :option:`--verbosity` for details.
+  string verbosity = 7;
+
+  string uri = 8;
+}
--- a/nighthawk/source/client/options_impl.cc
+++ b/nighthawk/source/client/options_impl.cc
@@ -1,0 +1,141 @@
+#include "nighthawk/source/client/options_impl.h"
+
+#include "tclap/CmdLine.h"
+
+#include "nighthawk/source/common/utility.h"
+
+namespace Nighthawk {
+namespace Client {
+
+OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
+  const char* descr = "Nighthawk, a L7 HTTP protocol family benchmarking tool based on Envoy.";
+  TCLAP::CmdLine cmd(descr, ' ', "PoC"); // NOLINT
+
+  TCLAP::ValueArg<uint64_t> requests_per_second("", "rps",
+                                                "The target requests-per-second rate. Default: 5.",
+                                                false, 5 /*default qps*/, "uint64_t", cmd);
+  TCLAP::ValueArg<uint64_t> connections(
+      "", "connections",
+      "The number of connections per event loop that the test should maximally "
+      "use. Default: 1.",
+      false, 1, "uint64_t", cmd);
+  TCLAP::ValueArg<uint64_t> duration("", "duration",
+                                     "The number of seconds that the test should run. Default: 5.",
+                                     false, 5, "uint64_t", cmd);
+  TCLAP::ValueArg<uint64_t> timeout(
+      "", "timeout",
+      "Timeout period in seconds used for both connection timeout and grace period waiting for "
+      "lagging responses to come in after the test run is done. Default: 5.",
+      false, 5, "uint64_t", cmd);
+
+  TCLAP::SwitchArg h2("", "h2", "Use HTTP/2", cmd);
+
+  TCLAP::ValueArg<std::string> concurrency(
+      "", "concurrency",
+      "The number of concurrent event loops that should be used. Specify 'auto' to let "
+      "Nighthawk leverage all vCPUs that have affinity to the Nighthawk process.Note that "
+      "increasing this results in an effective load multiplier combined with the configured-- rps "
+      "and --connections values.Default : 1. ",
+      false, "1", "string", cmd);
+
+  std::vector<std::string> log_levels;
+  log_levels.emplace_back("trace");
+  log_levels.emplace_back("debug");
+  log_levels.emplace_back("info");
+  log_levels.emplace_back("warn");
+  log_levels.emplace_back("error");
+  TCLAP::ValuesConstraint<std::string> verbosities_allowed(log_levels);
+
+  TCLAP::ValueArg<std::string> verbosity(
+      "v", "verbosity",
+      "Verbosity of the output. Possible values: [trace, debug, info, warn, error, critical]. The "
+      "default level is 'info'.",
+      false, "info", &verbosities_allowed, cmd);
+
+  TCLAP::UnlabeledValueArg<std::string> uri("uri",
+                                            "uri to benchmark. http:// and https:// are supported, "
+                                            "but in case of https no certificates are validated.",
+                                            true, "", "uri format", cmd);
+
+  cmd.setExceptionHandling(false);
+  try {
+    cmd.parse(argc, argv);
+  } catch (TCLAP::ArgException& e) {
+    try {
+      cmd.getOutput()->failure(cmd, e);
+    } catch (const TCLAP::ExitException&) {
+      // failure() has already written an informative message to stderr, so all that's left to do
+      // is throw our own exception with the original message.
+      throw MalformedArgvException(e.what());
+    }
+  } catch (const TCLAP::ExitException& e) {
+    // parse() throws an ExitException with status 0 after printing the output for --help and
+    // --version.
+    throw NoServingException();
+  }
+
+  requests_per_second_ = requests_per_second.getValue();
+  connections_ = connections.getValue();
+  duration_ = duration.getValue();
+  timeout_ = timeout.getValue();
+  uri_ = uri.getValue();
+  h2_ = h2.getValue();
+  concurrency_ = concurrency.getValue();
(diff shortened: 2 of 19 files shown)


oschaaf:
@htuch let me know if this PR size works for you

nighthawk/source/client/options.proto
```diff
@@ -0,0 +1,24 @@
...
+  // See :option:`--timeout` for details.
+  google.protobuf.Duration timeout = 4;
+  // See :option:`--h2` for details.
+  bool h2 = 5;
+  // See :option:`--concurrency` for details.
+  string concurrency = 6;
+  // See :option:`--verbosity` for details.
+  string verbosity = 7;
+
+  string uri = 8;
```
htuch: Nit: comment

nighthawk/source/client/options.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
+
+package nighthawk.client;
+
+import "google/protobuf/duration.proto";
+
+message CommandLineOptions {
```
htuch: Ultimately this will be a load test specification, the fact that it can arrive via CLI is just a concrete detail. Consider changing to reflect this or leave a TODO and we can do it later.

nighthawk/source/client/options_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+  requests_per_second_ = requests_per_second.getValue();
+  connections_ = connections.getValue();
+  duration_ = duration.getValue();
+  timeout_ = timeout.getValue();
+  uri_ = uri.getValue();
+  h2_ = h2.getValue();
+  concurrency_ = concurrency.getValue();
+  verbosity_ = verbosity.getValue();
+
+  uint64_t pretty_large_value = 1000 * 1000 * 1000;
```
htuch: Nit: const

nighthawk/source/client/options_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+  // We cap on negative values. TCLAP accepts negative values which we will get here as very
+  // large values. We just cap values to a very large maximum.
```
htuch: It seems we should either have per-option max (and put it in docs) or just allow anything that is representable in 2^63.

nighthawk/source/client/options_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+  if (requests_per_second_ == 0 || requests_per_second_ > pretty_large_value) {
+    throw MalformedArgvException("Invalid value for --requests_per_second");
+  }
+  if (connections_ == 0 || connections_ > pretty_large_value) {
+    throw MalformedArgvException("Invalid value for --connections");
```
htuch: I see you have tests for many of these, awesome. Would be nice to be able to see a coverage report to know we've got it at 100%.

nighthawk/source/client/utility.cc
```diff
@@ -0,0 +1,64 @@
...
+// returns 0 on failure. returns the number of HW CPU's
+// that the current thread has affinity with.
+// TODO(oschaaf): mull over what to do w/regard to hyperthreading.
+uint32_t determineCpuCoresWithAffinity() {
+  uint32_t concurrency = 0;
+  int i;
+  pthread_t thread = pthread_self();
+  cpu_set_t cpuset;
+  CPU_ZERO(&cpuset);
+  i = pthread_getaffinity_np(thread, sizeof(cpu_set_t), &cpuset);
```
htuch: One thing we should decide on early is if we want an OS syscall abstraction layer. It's pretty nice for when you want to mock, and isn't _that_ much additional boilerplate to use. See https://github.com/envoyproxy/envoy/blob/master/include/envoy/api/os_sys_calls.h and its uses in  Envoy.

nighthawk/source/client/utility.cc
```diff
@@ -0,0 +1,64 @@
...
+  if (i != 0) {
+    return 0;
+  } else {
+    for (i = 0; i < CPU_SETSIZE; i++) {
+      if (CPU_ISSET(i, &cpuset)) {
+        concurrency++;
```
htuch: Can you just use `CPU_COUNT` for the population count?

nighthawk/source/common/utility.h
```diff
@@ -0,0 +1,38 @@
...
+  Uri(std::string uri);
+
+  // TODO(oschaaf): username, password, etc. But we may want to look at
+  // pulling in a mature uri parser.
+  std::string host_and_port_;
+  std::string host_without_port_;
+  std::string path_;
+  uint64_t port_;
+  std::string scheme_;
+  bool is_https_;
```
htuch: It seems there is some redundancy in the strings above and things like schme_/is_https_. Seems like this is probably motivated by the needs of external code. Would it be possible to minimize this interface a bit?

nighthawk/source/client/utility.cc
```diff
@@ -0,0 +1,64 @@
+
+#include "absl/strings/match.h"
+#include "absl/strings/str_split.h"
+
+#include "common/http/utility.h"
+#include "common/network/utility.h"
+
+#include "nighthawk/source/common/utility.h"
+
+namespace Nighthawk {
```
htuch: Why is the `.cc` in `client` but the `.h` is in `common`?

nighthawk/source/common/rate_limiter_impl.cc
```diff
@@ -0,0 +1,32 @@
...
+}
+
+bool LinearRateLimiter::tryAcquireOne() {
+  if (acquireable_count_ > 0) {
+    acquireable_count_--;
+    acquired_count_++;
+    return true;
+  }
+
+  auto elapsed_since_start = time_source_.monotonicTime() - started_at_;
```
htuch: Nit: `const`

nighthawk/source/common/rate_limiter_impl.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "nighthawk/common/rate_limiter.h"
+
+#include "frequency.h"
+
+namespace Nighthawk {
+
+// Simple rate limiter that will allow acquiring at a linear pace.
+// The average rate is computed over a timeframe that starts at
+// instantiation.
```
htuch: Maybe comment on thread safety, as it's an open design question whether this should be shared or per-worker.

nighthawk/source/common/rate_limiter_impl.cc
```diff
@@ -0,0 +1,32 @@
...
+namespace Nighthawk {
+
+LinearRateLimiter::LinearRateLimiter(Envoy::TimeSource& time_source, const Frequency frequency)
+    : time_source_(time_source), acquireable_count_(0), acquired_count_(0), frequency_(frequency),
+      started_at_(time_source_.monotonicTime()) {
+  ASSERT(frequency.value() > 0, "Frequency must be > 0");
+}
+
+bool LinearRateLimiter::tryAcquireOne() {
+  if (acquireable_count_ > 0) {
```
htuch: I do worry about seeing bursty behavior still, even though we are now continuous and linear. This is more a non-actionable "something to think about", but I know from work on `httprr`, the interaction of thread scheduling and rate limiters like this can be quite jerky.

nighthawk/source/common/rate_limiter_impl.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "common/common/logger.h"
+#include "envoy/event/timer.h"
+
+#include "nighthawk/common/rate_limiter.h"
```
htuch: Re: header order, would be great to get a convention and have a check_format/fix_format script to enforce like on Envoy side.

nighthawk/source/common/utility.h
oschaaf: I removed `is_https_`; However, I put in `host_with_portname` and `host_without_portname` on purpose, to hopefully create awareness of that a distinct choice should be made here when using it. (In reaction to me messing up earlier by putting in the host without the portname in the `host:` header). Considering that, WDYT?

nighthawk/source/client/utility.cc
oschaaf: Whoops, moved the `.cc` to `common` so it lives next to the `.h`.

oschaaf:
@htuch I addressed most of the comments, and deferred some to the tracking issue over at  https://github.com/envoyproxy/envoy-perf/issues/32
PTAL

nighthawk/source/client/options_impl.cc
```diff
@@ -0,0 +1,144 @@
...
+
+  // concurrency must be either 'auto' or a positive integer.
+  if (concurrency_ != "auto") {
+    int parsed_concurrency;
+    try {
+      parsed_concurrency = std::stoi(concurrency_);
+    } catch (const std::invalid_argument& ia) {
+      throw MalformedArgvException("Invalid value for --concurrency");
+    } catch (const std::out_of_range& oor) {
+      throw MalformedArgvException("Invalid value for --concurrency");
```
htuch: Nit: the error message here could reflect "out of range", etc.

nighthawk/source/client/options_impl.cc
```diff
@@ -0,0 +1,144 @@
...
+    }
+    if (parsed_concurrency <= 0) {
+      throw MalformedArgvException("Invalid value for --concurrency");
```
htuch: Nit: `Invalid negative value for ...`

nighthawk/test/options_test.cc
```diff
@@ -0,0 +1,171 @@
...
+
+  std::string client_name_;
+  std::string good_test_uri_;
+  std::string no_arg_match_;
+};
+
+TEST_F(OptionsImplTest, BogusInput) {
+  // TODO(oschaaf): enable this test after we have host name validation.
+  // EXPECT_THROW_WITH_REGEX(createOptionsImpl(fmt::format("{} --foo", client_name_)),
+  //                        MalformedArgvException, no_arg_match_);
```
htuch: ?

nighthawk/test/options_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  EXPECT_EQ(cmd->timeout().seconds(), options->timeout().count());
+  EXPECT_EQ(cmd->h2(), options->h2());
+  EXPECT_EQ(cmd->concurrency(), options->concurrency());
+  EXPECT_EQ(cmd->verbosity(), options->verbosity());
+  EXPECT_EQ(cmd->uri(), options->uri());
+}
+
+// TODO(oschaaf): A couple of the tests below are pretty repetitive. They perform
+// the same sanity checking with just the argument name changing.
+// Figure out how to deduplicate.
```
htuch: Are these candidates for `TEST_P` parameterized tests?

nighthawk/test/utility_test.cc
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Nighthawk {
+
+class UtilityTest : public testing::Test {
+public:
+  UtilityTest() {}
+  void checkUriParsing(const std::string& uri_to_test, const std::string& host_and_port,
+                       const std::string& host_without_port, const uint64_t port,
+                       const std::string& scheme, const std::string& path) {
+    Uri uri = Uri::Parse(uri_to_test);
```
htuch: Nit: const

nighthawk/test/options_test.cc
oschaaf: Added a more elaborate comment here. 
```
  // TODO(oschaaf): enable this test after we have host name validation.
  // When just passing the non-existing argument --foo it would be interpreted as a
  // hostname. However, hostnames shouldn't start with '-', and hence this test should
  // not pass.
  // EXPECT_THROW_WITH_REGEX(createOptionsImpl(fmt::format("{} --foo", client_name_)),
  //                        MalformedArgvException, no_arg_match_);
```

nighthawk/test/options_test.cc
oschaaf: Yes; with `TEST_P` this looks much better, thanks. I planned to circle back to this myself but forgot to.

oschaaf:
@htuch thanks for taking another pass, comments addressed!

PR #48, opened 2019-02-14
Nighthawk: Statistics and updates to tooling & CI

oschaaf:
This adds a statistics interface / classes / tests,  hdr_histogram as a submodule, and a bunch of updates to `tools/` and the `.circleci/`.


ci/do_ci.sh
```diff
@@ -25,13 +37,16 @@ if [ -n "$CIRCLECI" ]; then
...
-    export BAZEL_TEST_OPTIONS="${BAZEL_TEST_OPTIONS} --local_resources=4096,2,1 --local_test_jobs=4"
+    export BAZEL_BUILD_OPTIONS="${BAZEL_BUILD_OPTIONS} --jobs 8"
+    export BAZEL_TEST_OPTIONS="${BAZEL_TEST_OPTIONS} --jobs 8 --local_test_jobs=8"
+    export MAKEFLAGS="-j 8"
 fi
 
-export PATH=/usr/lib/llvm-7/bin:$PATH
-export CC=clang
-export CXX=clang++
+if [ "$1" != "coverage" ]; then
```
htuch: You probably want to explicitly set `CC`/`CXX` to `gcc`/`g++` for coverage.

nighthawk/BUILD
```diff
@@ -1,10 +1,24 @@
...
     "@envoy//bazel:envoy_build_system.bzl",
     "envoy_cc_binary",
+    "envoy_cc_library",
 )
 
 package(default_visibility = ["//visibility:public"])
 
+envoy_cc_library(
+    name = "hdrhistogram_c",
+    srcs = glob(["hdrhistogram_c/src/*.c"]),
```
htuch: I'm not a huge fan of glob based builds of non-Bazel projects, but as long as you've verified each of these globbed files is supposed to be there (and not some test or utility), then this is fine.

nighthawk/BUILD
```diff
@@ -1,10 +1,24 @@
...
+    hdrs = glob(["hdrhistogram_c/src/*.h"]),
+    copts = [
+        "-std=gnu99",
+        "-Wno-implicit-function-declaration",
+        "-Wno-error",
+    ],
+    repository = "@envoy",
```
htuch: Shouldn't this be the hdrhistogram submodule repo? Or is it not a local repository and is just being overlaid on top of the main Envoy repo?

nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Only used in tests to match expectations to the right precision level.
+   * @return Number of significant digits. 0 is assumed to be max precision.
+   */
+  virtual uint64_t significant_digits() const { return 0; }
+
+  /**
+   * Dumps a representation of the statistic in plain text to stdout.
+   */
+  virtual void dumpToStdOut(const std::string& header) const PURE;
```
htuch: I think converting to string would be a more flexible API, then it's easy to dump that at the caller to stdout.

nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Updates the proto output to reflect the contents of the statistic.
+   */
+  virtual void toProtoOutput(nighthawk::client::Output& output) PURE;
+
+  /**
+   * Combines two Statistics into one, and returns a new, merged, Statistic.
+   * This is useful for computing results from multiple workers into a
+   * single global view.
+   * @param a The Statistic that should be combined with this instance.
```
htuch: Nit: prefer a better arg name than `a`

nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -0,0 +1,57 @@
...
+  virtual uint64_t count() const PURE;
+  virtual double mean() const PURE;
+  virtual double variance() const PURE;
+  virtual double stdev() const PURE;
+
+  /**
+   * Only used in tests to match expectations to the right precision level.
+   * @return Number of significant digits. 0 is assumed to be max precision.
+   */
+  virtual uint64_t significant_digits() const { return 0; }
```
htuch: Nit: on an interface I'd use `significantDigits()`

nighthawk/source/client/output.proto
```diff
@@ -0,0 +1,25 @@
...
+import "google/protobuf/timestamp.proto";
+
+import "nighthawk/source/client/options.proto";
+
+
+message Percentile {
+    google.protobuf.Duration latency = 1;
+    double percentile = 2;
+    int64 count = 3;
+    double inverted_percentile = 4;
```
htuch: Can you add comments to these? Mostly self-explanatory, but inverted percentile isn't..

nighthawk/source/client/output.proto
```diff
@@ -0,0 +1,25 @@
...
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+import "nighthawk/source/client/options.proto";
+
+
+message Percentile {
+    google.protobuf.Duration latency = 1;
+    double percentile = 2;
+    int64 count = 3;
```
htuch: Nit: prefer unsigned here.

nighthawk/source/client/output.proto
```diff
@@ -0,0 +1,25 @@
...
+    double inverted_percentile = 4;
+}
+
+message Output {
+  google.protobuf.Timestamp timestamp= 1;
```
htuch: Super nit: `timestamp = 1`

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,212 @@
...
+void StatisticImpl::dumpToStdOut(const std::string& header) const {
+  ENVOY_LOG(info, "{}", header);
+  ENVOY_LOG(info, "#Completed: {}. Mean: {:.{}f}μs. Stdev: {:.{}f}μs.", count(), mean() / 1000, 2,
+            stdev() / 1000, 2);
+}
+
+void StatisticImpl::toProtoOutput(nighthawk::client::Output& output) {
+  output.set_request_count(count());
+  output.mutable_mean()->set_nanos(mean());
+  output.mutable_stdev()->set_nanos(stdev());
```
htuch: No percentiles?

nighthawk/BUILD
oschaaf: It is being overlaid on top of the Envoy repo, without doing that I get `no such package 'bazel': BUILD file not found on package path and referenced by '//nighth
awk:hdrhistogram_c`

nighthawk/source/client/output.proto
oschaaf: I removed inverted_percentile, which actually just was `1/(1-Percentile)` and therefore redundant.

nighthawk/source/common/statistic_impl.cc
oschaaf: `StatisticImpl` is a base class for the other implementations. `HdrStatistic` is currently the only one that provides percentiles and overrides this to do so.

nighthawk/include/nighthawk/common/statistic.h
oschaaf: Agreed, done

nighthawk/BUILD
htuch: Yeah, TBH this is a pretty weird way to manage external repositories. I think for HDR histogram, since we don't need the regular ability to instrument and hack that tree, it should not be modeled as a submodule. Instead, I think we should make this an external repository in the Bazel sense and inject a BUILD file into it.

For reference, here's the way we've done this in the main Envoy repository in an analogous situation https://github.com/envoyproxy/envoy/blob/84d038f5a67a8dc8ca9753e9089a33d8110db3c2/api/bazel/repositories.bzl#L34 (and other examples in that file). Basically, make it a `http_archive` and set `build_file_content`. Then you can reference it by `@whatever_name_you_set` in the `name` field.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+
+StreamingStatistic::StreamingStatistic() : count_(0), mean_(0), sum_of_squares_(0) {}
+
+void StreamingStatistic::addValue(int64_t value) {
+  double delta, delta_n;
+  count_++;
+  delta = value - mean_;
+  delta_n = delta / count_;
+  mean_ += delta_n;
+  sum_of_squares_ += delta * delta_n * (count_ - 1);
```
htuch: This math seems a bit confusing; should `sum_of_squares_` be literally the sum of the squares of `value`? Then we do the heavy lifting in `variance()` like in https://stackoverflow.com/a/1175084.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+std::unique_ptr<Statistic> StreamingStatistic::combine(const Statistic& statistic) {
+  const StreamingStatistic& a = *this;
+  const StreamingStatistic& b = dynamic_cast<const StreamingStatistic&>(statistic);
+  auto combined = std::make_unique<StreamingStatistic>();
+
+  combined->count_ = a.count() + b.count();
+  combined->mean_ = ((a.count() * a.mean()) + (b.count() * b.mean())) / combined->count_;
+  combined->sum_of_squares_ =
+      a.sum_of_squares_ + b.sum_of_squares_ +
+      pow(a.mean() - b.mean(), 2) * a.count() * b.count() / combined->count();
```
htuch: Also wondering if this math can be simplified.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+  auto combined = std::make_unique<InMemoryStatistic>();
+  const InMemoryStatistic& b = dynamic_cast<const InMemoryStatistic&>(statistic);
+
+  combined->samples_.insert(combined->samples_.end(), this->samples_.begin(), this->samples_.end());
+  combined->samples_.insert(combined->samples_.end(), b.samples_.begin(), b.samples_.end());
+  combined->streaming_stats_ = this->streaming_stats_->combine(*b.streaming_stats_);
+  return combined;
+}
+
+const int HdrStatistic::SIGNIFICANT_DIGITS = 4;
```
htuch: Nit: probably should call this `SignificantDigits`

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+
+HdrStatistic::HdrStatistic() : histogram_(nullptr) {
+  // Upper bound of 60 seconds (tracking in nanoseconds).
+  const uint64_t max_latency = 1000L * 1000 * 1000 * 60;
+
+  int status = hdr_init(1 /* min trackable value */, max_latency, HdrStatistic::SIGNIFICANT_DIGITS,
+                        &histogram_);
+  if (status != 0) {
+    ENVOY_LOG(error, "Failed to initialize HdrHistogram.");
+    histogram_ = nullptr;
```
htuch: I wonder if this should be a `throw` and we crash out? Seems like an unlikely scenario.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+  if (histogram_ != nullptr) {
+    hdr_close(histogram_);
+    histogram_ = nullptr;
+  }
+}
+
+void HdrStatistic::addValue(int64_t value) {
+  if (histogram_ != nullptr) {
+    if (!hdr_record_value(histogram_, value)) {
+      ENVOY_LOG(warn, "Failed to record value into HdrHistogram.");
```
htuch: When/how does this happen?

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+  ;
+}
+double HdrStatistic::stdev() const {
+  // HdrHistogram_c's stdev actually gives us the population standard deviation.
+  // So we compute the sample standard deviation ourselves instead.
+  // TODO(oschaaf): this fixes some of the test expectations, but figure out if
+  // stdev or pstdev is preferrable. Looks like wrk2 uses pstdev which would produce
+  // (slightly) better numbers, though that probably isn't a reason for us to decice
+  // which one to use here. Switching to pstdev would get rid of having to do this
+  // ourselves.
```
htuch: When measuring latency of a real system, i.e. just probing, arguably we are just sampling. OTOH, when doing load tests, all the traffic is from the load generator, so could be thought of as the population. Not sure it's hugely important though as long as we document it, we're just after some idea of variance. If we can simplify the below, it sounds like a win.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+
+  if (this->histogram_ == nullptr || b.histogram_ == nullptr) {
+    return combined;
+  }
+
+  int dropped;
+  dropped = hdr_add(combined->histogram_, this->histogram_);
+  dropped += hdr_add(combined->histogram_, b.histogram_);
+  if (dropped > 0) {
+    ENVOY_LOG(warn, "Combining HdrHistograms dropped values.");
```
htuch: When does this happen? Should we abort?

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+  }
+  return combined;
+}
+
+std::unique_ptr<HdrStatistic> HdrStatistic::getCorrected(const Frequency& frequency) {
```
htuch: I'm still curious about the math and algorithm behind this correction for coordinated omission? Do you have references? Maybe add them as comments?

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,216 @@
...
+    ENVOY_LOG(warn, "HdrHistogram latencies could not be printed.");
+    return stream.str();
+  }
+
+  stream << fmt::format("{:>12} {:>14} (us)", "Percentile", "Latency") << std::endl;
+
+  std::vector<double> percentiles{50.0, 75.0, 90.0, 99.0, 99.9, 99.99, 99.999, 100.0};
+  for (uint64_t i = 0; i < percentiles.size(); i++) {
+    double p = percentiles[i];
+    int64_t n = hdr_value_at_percentile(histogram_, p);
```
htuch: Nit: `const` these

nighthawk/source/common/statistic_impl.h
```diff
@@ -0,0 +1,93 @@
...
+  uint64_t count() const override;
+  double mean() const override;
+  double variance() const override;
+  double stdev() const override;
+
+  std::unique_ptr<Statistic> combine(const Statistic& statistic) override;
+
+  /**
+   * Gets a HdrStatistic instance with corrections for coordinated omission applied.
+   * This should be done only once.
```
htuch: Why only once? This seems a fragile interface.

nighthawk/source/common/statistic_impl.h
```diff
@@ -0,0 +1,93 @@
...
+   * @param frequency The frequency at which latencies ought to have been measured.
+   * NOTE: this is neither used nor tested at the moment.
```
htuch: I'm voting to maybe drop it for now and introduce it later as needed.

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,85 @@
...
+public:
+  /**
+   * With 0 significant digits passed, this uses EXPECT_DOUBLE_EQ. Otherwise EXPECT_NEAR
+   * will be called with a computed acceptable range based on the number of significant
+   * digits and tested_value.
+   * @param expected_value the expected value
+   * @param tested_value the tested_value
+   * @param significant the number of significant digits that should be used to compare values.
+   */
+  static void expect_near(double expected_value, double tested_value, uint64_t significant) {
```
htuch: Nit: `expectNear`

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,85 @@
...
+    if (significant > 0) {
+      EXPECT_NEAR(expected_value, tested_value,
+                  std::pow(10, std::ceil(std::log10(tested_value)) - 1 - significant));
+    } else {
+      EXPECT_DOUBLE_EQ(expected_value, tested_value);
+    }
+  }
+
+private:
+  Helper() = default;
```
htuch: Nit: not needed

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,85 @@
...
+};
+
+TYPED_TEST_SUITE(TypedStatisticTest, MyTypes);
```
htuch: What might be cool is a test that runs all three types with the same data and compares their outputs! :)

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,85 @@
...
+
+  Helper::expect_near(2295675, b.mean(), a.significantDigits());
+  Helper::expect_near(13561820041021, b.variance(), a.significantDigits());
+  Helper::expect_near(3682637.6472605884, b.stdev(), a.significantDigits());
+
+  auto c = a.combine(b);
+  EXPECT_EQ(6, c->count());
+  Helper::expect_near(1147838.5, c->mean(), c->significantDigits());
+  Helper::expect_near(7005762373287.5, c->variance(), c->significantDigits());
+  Helper::expect_near(2646840.0732359141, c->stdev(), c->significantDigits());
```
htuch: Probably should add tests for other methods, like `toProtoOutput` etc.

nighthawk/source/common/statistic_impl.cc
oschaaf: Added a comment:
```c++
    // Failure to record a value can happen when it exceeds the configured minimum
    // or maximum value we passed when initializing histogram_.
```

nighthawk/source/common/statistic_impl.cc
oschaaf: Added a comment here as well to clarify when this may happen. We could abort. Or maybe it's nice to set a flag on the Statistic that an event occurred that may influence reporting accuracy. We can use that in the proto output, and this warning should show up when the CLI is used. (maybe we should direct messages leveled > warning to the proto output as well?).

nighthawk/source/common/statistic_impl.cc
oschaaf: I agree with removing this for now. The correction applied is based on the actual frequency vs the expected frequency; in the quick tests I ran with it I don't think it is of any use to us. It's probably better to track time spend blocked in closed-loop mode ourselves and report a second Statistic (or histogram) based on that somewhere in the future.

nighthawk/source/common/statistic_impl.h
oschaaf: You would be correcting values that are already corrected, which would be no good according to hdr_histogram docs. We could guard against that.. But it's probably not relevant, because I will remove this for now per your comment below.

nighthawk/source/common/statistic_impl.cc
oschaaf: I (tried to) implement knuths way of computing the running stats here, which should be numerically stable and reasonably accurate. The output looks good in tests; Maybe defer this to the track-for-future list for revisiting later for now? Maybe it would be nice to add test designed for catching numerical stability issues as well..

oschaaf: ps - I switched everything to the population standard dev, which slightly simplifies this, and allows us to use a call from hdrhistogram_c, dropping some more complex code in the `HdrStatistic` implementation.

oschaaf: Added a test for catastrophic cancellation, plus an implementation of the algorithm you pointed out. Added notes in the comments, see: https://github.com/envoyproxy/envoy-perf/pull/48/files#diff-653bc7d320cfd8bc55e04f79b81e5f0eR93
(side-note: the values used in the test for this proves variance will return negative values with the newly added SimpleStats implementation, which I think is an effect that  may have caused the 'negative sigma' errors when we took a look at fortio a while back).

ci/do_ci.sh
```diff
@@ -1,18 +1,30 @@
...
+}
+
 function do_clang_tidy() {
     ci/run_clang_tidy.sh
 }
 
+function do_coverage() {
+    gcc -v && \
+    gcov -v && \
+    bazel coverage $BAZEL_BUILD_OPTIONS --test_output=all --verbose_failures --experimental_cc_coverage nighthawk/test/...  --instrumentation_filter=//nighthawk/...,. --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main --combined_report=lcov; genhtml bazel-out/_coverage/_coverage_report.dat
```
htuch: This should be `BAZEL_TEST_OPTIONS`, since coverage runs the tests.

nighthawk/BUILD
htuch: Do you want to sort this one out in this PR or in a followup PR?

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,140 @@
...
+
+  a.addValue(2);
+  EXPECT_EQ(2, a.count());
+  EXPECT_DOUBLE_EQ(1.5, a.mean());
+  EXPECT_DOUBLE_EQ(0.25, a.pvariance());
+  EXPECT_DOUBLE_EQ(0.5, a.pstdev());
+}
+
+TYPED_TEST(TypedStatisticTest, CatastrophicalCancellation) {
+  // From https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
```
htuch: Yeah, this cancellation property seems bad. I think it's fine to just use a one-pass algorithm that can avoid cancelation as described in the Wiki, we just need to provide a good reference and explain why this was selected.

nighthawk/source/common/statistic_impl.cc
htuch: Yeah, it would be good to avoid the negative sigma possibility. It's fine using this algorithm, just providing good references to where it comes from would help the reader. Also, I think `sum_of_squares_` is not quite the right name, it's an aggregation of square distances, but there's a bit more going on than just sum of squares.

nighthawk/source/common/statistic_impl.cc
htuch: Can't we configure buckets for underflow/overflow? In `httprr`, I think we had something like this.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,214 @@
...
+    return 0;
+  }
+  return hdr_stddev(histogram_);
+}
+
+std::unique_ptr<Statistic> HdrStatistic::combine(const Statistic& statistic) {
+  auto combined = std::make_unique<HdrStatistic>();
+  const HdrStatistic& b = dynamic_cast<const HdrStatistic&>(statistic);
+
+  if (this->histogram_ == nullptr || b.histogram_ == nullptr) {
```
htuch: Nit: Shouldn't need `this_`.

nighthawk/source/common/statistic_impl.cc
```diff
@@ -0,0 +1,214 @@
...
+    return combined;
```
htuch: I think it would majorly simplify if we just `throw` in the constructor when `nullptr` happens, so we don't need special case handling around the place.

nighthawk/source/common/statistic_impl.cc
htuch: Yeah, I have a comment about about whether we can have buckets for under/overflow..

nighthawk/test/statistic_test.cc
htuch: Also, maybe throwing in some some different pseudo-random examples would be nice bonus. This seems the sort of thing we can mechanically verify across implementations without a lot of manual test creation.

nighthawk/BUILD
oschaaf: Moved this to the track-for-future list, I prefer doing it in a follow up.

oschaaf:
@htuch modulo the remark about  "buckets for under/overflow.." for hdrhistogram_c I think I addressed everything. Do you have a reference for that? You have to specify the minimum and maximum values that you anticipate to process at creation time, e.g.:

```c
// Initialise the histogram
hdr_init(
    1,  // Minimum value
    INT64_C(3600000000),  // Maximum value
    3,  // Number of significant figures
    &histogram)  // Pointer to initialise
```

As I understand it, values out of range will just be dropped.

htuch:
@oschaaf it's fine if HDR histogram doesn't natively support this. In histogram implementations (admittedly much more naive) that I've built in the past, adding buckets for underflow/overflow were pretty straightforward.

nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -0,0 +1,71 @@
...
+   */
+  virtual nighthawk::client::Statistic toProto() PURE;
+
+  /**
+   * Combines two Statistics into one, and returns a new, merged, Statistic.
+   * This is useful for computing results from multiple workers into a
+   * single global view. Types of the Statistics objects that will be combined
+   * must be the same, or else a std::bad_cast exception will be raised.
+   * @param statistic The Statistic that should be combined with this instance.
+   * @return T Merged Statistic instance.
```
htuch: Nit: `@return std::unique_ptr<Statistic> instance`.

nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -0,0 +1,71 @@
...
+   */
+  virtual bool resistsCatastrophicCancellation() const { return false; }
+
+  /**
+   * Gets a representation of the statistic as a std::string.
+   */
+  virtual std::string toString() const PURE;
+
+  /**
+   * Gets a proto Output reflecting the contents of the statistic.
```
htuch: Nit: s/Output/outputs/

ci/do_ci.sh
```diff
@@ -25,13 +37,19 @@ if [ -n "$CIRCLECI" ]; then
...
 fi
 
-export PATH=/usr/lib/llvm-7/bin:$PATH
-export CC=clang
-export CXX=clang++
+if [ "$1" != "coverage" ]; then
+    export PATH=/usr/lib/llvm-7/bin:$PATH
+    export CC=clang
+    export CXX=clang++
+else
```
htuch: Nit: can you flip the logic here so we have `if coverage then .. else ..`, it's easier to parse.

nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Indicates if the implementation is subject to catastrophic cancellation.
+   * Used in tests.
+   * @return True iff catastrophic cancellation should not occur.
+   */
+  virtual bool resistsCatastrophicCancellation() const { return false; }
+
+  /**
+   * Gets a representation of the statistic as a std::string.
```
htuch: Nit: prefer one liner `@return std::string a representation of ..` style for these.

nighthawk/source/common/statistic_impl.h
```diff
@@ -0,0 +1,111 @@
...
+private:
+  uint64_t count_;
+  double sum_x_;
+  double sum_x2_;
+};
+
+/**
+ * Statistic that keeps track of count/mean/pvariance/pstdev with low memory
+ * requirements. Resistant to catastrophic cancellation and pretty accurate.
+ * Based on Donald Knuth's online variance computation algorithm.
```
htuch: Can you add a more specific reference? URL or book/paper/page number etc.

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,199 @@
...
+    // TODO(oschaaf): evaluate ^^
+    if (a.resistsCatastrophicCancellation()) {
+      Helper::expectNear(22.5, a.pvariance(), a.significantDigits());
+      Helper::expectNear(4.7434164902525691, a.pstdev(), a.significantDigits());
+    }
+  }
+}
+
+TYPED_TEST(TypedStatisticTest, OneMillionRandomSamples) {
+  std::random_device rd;
```
htuch: Can you make this PRNG? The problem with non-deterministic generation is we can't replicate failures.

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,199 @@
...
+
+  for (int i = 1; i <= max; i++) {
+    statistic.addValue(i);
+  }
+
+  std::string str;
+  google::protobuf::util::JsonPrintOptions options;
+  google::protobuf::util::MessageToJsonString(statistic.toProto(), &str, options);
+  std::ifstream myfile;
+  myfile.open("nighthawk/test/hdr_proto_json.gold");
```
htuch: Take a look at `TestEnvironment::runfilesPath()` in the main Envoy repo; can we use this here? Or some similar abstraction? The main concern I have is that when we import into Google and make use of this internally, this path calculation changes.

oschaaf:
@htuch two questions:
- The test with the `.gold` file now uses `Envoy::TestEnvironment::runfilesPath`. Can you verify this now works OK in your environment?
- Separately bucketing values that are out-of-range for  `hdrhistogram_c`: to make sure I understand correctly, no action required on this (at least for now), correct?

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,209 @@
...
+
+  std::string str;
+  google::protobuf::util::JsonPrintOptions options;
+  google::protobuf::util::MessageToJsonString(statistic.toProto(), &str, options);
+  std::ifstream myfile;
+  // TODO(oschaaf): If this is confirmed to work in Google's environment, wrap
+  // Envoy::TestEnvironment::runfilesPath in a call within the Nighthawk namespace for re-use.
+  // TODO(oschaaf): Set up reusable helpers for .gold file testing, we'll probably get more of that.
+  // Also, move out the test data into its own directory.
+  myfile.open(Envoy::TestEnvironment::runfilesPath("nighthawk/test/hdr_proto_json.gold"));
```
htuch: Nit: you can also reuse https://github.com/envoyproxy/envoy/blob/master/source/common/filesystem/filesystem_impl.h#L52.

nighthawk/test/statistic_test.cc
```diff
@@ -0,0 +1,209 @@
...
+  EXPECT_EQ(proto.count(), 2);
+  EXPECT_EQ(proto.mean().nanos(), std::round(a.mean()));
+  EXPECT_EQ(proto.pstdev().nanos(), std::round(a.pstdev()));
+}
+
+class StatisticTest : public testing::Test {};
+
+TEST(StatisticTest, HdrStatisticPercentilesProto) {
+  HdrStatistic statistic;
+  int max = 10;
```
htuch: Nit: I'd just inline this.

nighthawk/test/statistic_test.cc
```diff
@@ -173,25 +179,21 @@ TYPED_TEST(TypedStatisticTest, ProtoOutput) {
...
-  // TODO(oschaaf): If this is confirmed to work in Google's environment, wrap
-  // Envoy::TestEnvironment::runfilesPath in a call within the Nighthawk namespace for re-use.
-  // TODO(oschaaf): Set up reusable helpers for .gold file testing, we'll probably get more of that.
-  // Also, move out the test data into its own directory.
-  myfile.open(Envoy::TestEnvironment::runfilesPath("nighthawk/test/hdr_proto_json.gold"));
-  std::string gold;
-  myfile >> gold;
-  EXPECT_EQ(gold, str);
+  Envoy::MessageUtil util;
+  util.loadFromJson(filesystem.fileReadToEnd(TestEnvironment::runfilesPath(
```
htuch: Hmm, my suggested changes made this look worse, they were bad. I think we should have an explicit equality method, rather than implicit call on `MessageUtil`, but that change needs to happen upstream. Can you do this? I also didn't realize we needed a full filesystem song-and-dance, I thought we could just use a static method on the class. I'll merge this PR, but please do consider undoing my bad advice in a future PR.

oschaaf: I'll take a look at upstreaming that change and incorporating that here.

PR #49, opened 2019-02-19
 Sequencer interface, implementation & tests 

oschaaf:


nighthawk/include/nighthawk/common/sequencer.h
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Abstract Sequencer interface. The Sequencer will drive calls to the SequencerTarget.
+ * The contract with the target is that it will call the provided callback when it is ready.
+ * The target will return true if it was able to proceed, or false if a retry is warranted at
+ * a later time (because of being out of required resources, for example).
+ */
+class Sequencer {
+public:
+  Sequencer(SequencerTarget& target) : target_(target) {}
```
htuch: This seems like implementation in the abstract interface.

nighthawk/include/nighthawk/common/sequencer.h
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Wait until the sequencer has finished.
+   */
+  virtual void waitForCompletion() PURE;
+
+  /**
+   * @return double an up-to-date completions per second rate.
+   */
+  virtual double completions_per_second() PURE;
```
htuch: Nit: prefer `completionsPerSecond()` (and same below) for abstract interface methods. Also, some of these could probably be `const`, since they shouldn't need to mutate, e.g. `virtual double completionsPerSecond() const PURE;`

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,113 @@
...
+
+#include "common/common/assert.h"
+
+#include "nighthawk/source/common/sequencer_impl.h"
+
+using namespace std::chrono_literals;
+
+namespace Nighthawk {
+
+const std::chrono::milliseconds SequencerImpl::EnvoyTimerMinResolution = 1ms;
```
htuch: `constexpr`?

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,113 @@
...
+SequencerImpl::SequencerImpl(Envoy::Event::Dispatcher& dispatcher, Envoy::TimeSource& time_source,
+                             RateLimiter& rate_limiter, SequencerTarget& target,
+                             std::chrono::microseconds duration,
+                             std::chrono::microseconds grace_timeout)
+    : Sequencer(target), dispatcher_(dispatcher), time_source_(time_source),
+      rate_limiter_(rate_limiter), duration_(duration), grace_timeout_(grace_timeout),
+      start_(time_source.monotonicTime().min()), targets_initiated_(0), targets_completed_(0),
+      spin_when_idle_(true) {
+  if (target_ == nullptr) {
+    throw NighthawkException("SequencerImpl must be constructed with a SequencerTarget.");
```
htuch: I think if `target` is a `&`, it's reasonable to assume non-null.

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,113 @@
...
+  incidental_timer_->disableTimer();
+  periodic_timer_.reset(nullptr);
+  incidental_timer_.reset(nullptr);
+  dispatcher_.exit();
+}
+
+void SequencerImpl::run(bool from_timer) {
+  const auto now = time_source_.monotonicTime();
+
+  if ((now - start_) > (duration_)) {
```
htuch: Can you add some comments explaining the flow of this method? E.g. tell the narrative "Now we check for X, then we check for Y then we do Z..".

nighthawk/source/common/sequencer_impl.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+  SequencerImpl(Envoy::Event::Dispatcher& dispatcher, Envoy::TimeSource& time_source,
+                RateLimiter& rate_limiter, SequencerTarget& target,
+                std::chrono::microseconds duration, std::chrono::microseconds grace_timeout);
+  void start() override;
+  void waitForCompletion() override;
+
+  // TODO(oschaaf): calling this after stop() will return broken/unexpected results.
+  double completions_per_second() override {
+    double us =
```
htuch: Nit: `const`

nighthawk/source/common/sequencer_impl.h
```diff
@@ -0,0 +1,65 @@
...
+    return us == 0 ? 0 : ((targets_completed_ / us) * 1000000);
+  }
+
+  // TODO(oschaaf): we want to track time we wait between the rate limiter
+  // indicating we should call target_() but target returs false, meaning
+  // we are blocked / have entered closed loop mode.
+  const HdrStatistic& blocked_statistic() override { return blocked_statistic_; }
+  const HdrStatistic& latency_statistic() override { return latency_statistic_; }
+
+  // Spinning makes tests hang when simulated time is used.
```
htuch: Can't we avoid this by mocking out the `pthread_yield` using the `Api` layer in Envoy? This would allow us to increment simulated time.

nighthawk/test/sequencer_test.cc
```diff
@@ -0,0 +1,108 @@
...
+  Envoy::Thread::ThreadFactoryImplPosix thread_factory_;
+  Envoy::Stats::IsolatedStoreImpl store_;
+  Envoy::Event::SimulatedTimeSystem time_system_;
+  Envoy::Api::Impl api_;
+  Envoy::Event::DispatcherPtr dispatcher_;
+  int callback_test_count_;
+  const Frequency frequency_;
+  const std::chrono::milliseconds interval_;
+  const uint64_t times_;
+  LinearRateLimiter rate_limiter_;
```
htuch: Should we have a `MockRateLimiter`? Would mocking `dispacher_` provide any value?

oschaaf: We now have `MockRateLimiter` and `MockPlatformUtils`. I didn't mock the `Dispatcher` yet (or use the existing mock for that)..

oschaaf:
@htuch I worked on addressing your feedback, I feel this is ready for another pass

nighthawk/test/mocks.h
```diff
@@ -0,0 +1,75 @@
...
+  MockPlatformUtil() { DelegateToSimulatedTimeAwarePlatformUtil(); }
+
+  MOCK_CONST_METHOD0(yieldCurrentThread, void());
+
+  void setTimeSystem(Envoy::Event::SimulatedTimeSystem& time_system) {
+    simulated_time_aware_platform_util_.setTimeSystem(time_system);
+  }
+
+private:
+  void DelegateToSimulatedTimeAwarePlatformUtil() {
```
htuch: Nit: `delegateToSim..`

nighthawk/test/mocks.h
```diff
@@ -0,0 +1,75 @@
...
+    // When this is called we are in a tight spin loop. SimulatedTimeAwarePlatformUtil moves the
+    // simulated time forward, avoiding the tests hanging.
+    ON_CALL(*this, yieldCurrentThread())
+        .WillByDefault(testing::Invoke(&simulated_time_aware_platform_util_,
+                                       &SimulatedTimeAwarePlatformUtil::yieldCurrentThread));
+  }
+
+  SimulatedTimeAwarePlatformUtil simulated_time_aware_platform_util_;
```
htuch: Why the extra level of delegation?

oschaaf: See the comment above?:
```
    // When this is called we are in a tight spin loop. SimulatedTimeAwarePlatformUtil moves the
    // simulated time forward, avoiding the tests hanging.
```

nighthawk/test/mocks.h
```diff
@@ -0,0 +1,75 @@
...
+    time_system_ = &time_system;
+  }
+
+private:
+  Envoy::Event::SimulatedTimeSystem* time_system_;
+};
+
+class MockPlatformUtil : public PlatformUtil {
+public:
+  MockPlatformUtil() { DelegateToSimulatedTimeAwarePlatformUtil(); }
```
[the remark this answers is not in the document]
oschaaf: Added to track-for-future

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,145 @@
...
+  } else {
+    // Re-schedule the periodic timer if it was responsible for waking up this code.
+    scheduleRun();
+  }
+}
+
+void SequencerImpl::waitForCompletion() {
+  ASSERT(running_);
+  dispatcher_.run(Envoy::Event::Dispatcher::RunType::Block);
+  if (running_) {
```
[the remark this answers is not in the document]
oschaaf: Good question, the only situation where `running_` equals `true` when we get here, is actually in one of the tests using mocks...
I need to think about this can actually really happen under normal conditions, maybe in an edge case, but if not it should probably be better to have `ASSERT(!running)` here.

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,145 @@
...
+SequencerImpl::SequencerImpl(PlatformUtil& platform_util, Envoy::Event::Dispatcher& dispatcher,
+                             Envoy::TimeSource& time_source, RateLimiter& rate_limiter,
+                             SequencerTarget& target, std::chrono::microseconds duration,
+                             std::chrono::microseconds grace_timeout)
+    : target_(target), platform_util_(platform_util), dispatcher_(dispatcher),
+      time_source_(time_source), rate_limiter_(rate_limiter), duration_(duration),
+      grace_timeout_(grace_timeout), start_(time_source.monotonicTime().min()),
+      targets_initiated_(0), targets_completed_(0), running_(false), blocked_(false) {
+  ASSERT(target_ != nullptr, "No SequencerTarget");
+  periodic_timer_ = dispatcher_.createTimer([this]() { run(true); });
```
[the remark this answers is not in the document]
oschaaf: Well, the periodic timer gets re-enabled by `run()`, and the adhoc timer does not. We're not *always* busy-looping, with long-running transactions the periodic timer is mostly driving this, and gives us a minimum of timely initiating new work. Tangentially related, I think that if we could somehow prioritize i/o events over timer events in libevent when we schedule a timer event to fire immediately, then I think we could always busy-loop and thus drop the periodic timer.

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,145 @@
...
+  adhoc_timer_ = dispatcher_.createTimer([this]() { run(false); });
+}
+
+SequencerImpl::~SequencerImpl() {}
+
+void SequencerImpl::start() {
+  ASSERT(!running_);
```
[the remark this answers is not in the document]
oschaaf: Right now it is actually a one shot sequencer. Are you suggesting renaming the class to something like `OneShotSequencerImpl`?

oschaaf:
@htuch Thanks! Resolved everything that I could address in the commit I just pushed, and left a few questions.

nighthawk/test/mocks.h
htuch: Why not just inject `SimulatedTimeAwarePlatform` directly?

nighthawk/test/mocks.h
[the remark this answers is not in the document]
htuch: Sure. Can you schedule some bandwidth to burn down some of these once we finish the upstreaming of the MVP?

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
htuch: My main concern is around complexity. Questions:
1. For the periodic timer, are we making a latency tradeoff? Is the worst outcome that the release of requests is a bit off, but our latency measurement once released is not affected?
2. Can you write up a detailed description of this interplay somewhere? It's kind of confusing to the reader what these are.
3. Would it make sense to rename `adhoc_timer_` as `wait_for_ratelimiter_timer_`?

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
htuch: Not rename, but make clear in all interface documents that this is one-shot only. Possibly validate with `ASSERT`s where it makes sense as well.

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
htuch: Yeah, anything to simplify reasoning has my +1.

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
oschaaf: 1. Timing is way more accurate

nighthawk/test/mocks.h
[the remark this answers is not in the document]
oschaaf: Re: burning down things from track-for-future once we finish the upstreaming of the MVP: sure thing! Already did some of them in time spend between posting updates here and waiting for review.

nighthawk/test/mocks.h
oschaaf: I'm not 100% sure I understand.. I rewrote this a bit, please check. I would like to be able to `EXPECT_CALL` on this, we use that in some of the sequencer tests:

```
  // We will be stepping in time at precisely the rate limiter frequency below.
  // As the callbacks won't complete, the sequencer will never consider itself idle.
  // Hence, no spinning, and no calls to yieldCurrentThread.
  EXPECT_CALL(platform_util_, yieldCurrentThread()).Times(0);
```

If there is a better way, can you post an example?

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
oschaaf: One more thing I wanted to add:
Observationally, it looks like the spin loop has a very positive effect on some systems. The laptop I am commenting with right now is not tuned for latency and has a i7-8750H cpu. 
Mean latency reported by wrk2 in a 1000rps test is 299usec, whereas NH reports 115 usec.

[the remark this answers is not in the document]
oschaaf: PS: Experimentally removing the spin loop, NH reports a mean of 336 usec.

nighthawk/test/mocks.h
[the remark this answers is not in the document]
oschaaf: Re: moved all constructors/destructors into `mocks.cc`, let me know if this looks better.

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
htuch: Yeah, this is pretty interesting. Maybe we want two modes of operation, one in which we always spin, and another in which we only spin when waiting for the rate limiter. I think "always spin" might be interesting for two reasons:
1. If the rate limiter is bursty at the libevent timer granularity, always spinning would allow us to achieve the finest granularity possible.
2. We would avoid having any CPU or OS level power management kick-in, by always maintaining the load generator in maximum C state.
This is of course, completely counter-intuitive from an energy efficiency or shared compute environment perspective, but I think it makes a lot of sense for load generation and latency measurement.

> Perhaps rename `adhoc_timer` to `wait_for_next_timer_`?

Sure.

nighthawk/test/mocks.h
```diff
@@ -0,0 +1,68 @@
...
+using namespace std::chrono_literals;
+
+constexpr std::chrono::milliseconds TimeResolution = 10ms;
+
+namespace Nighthawk {
+
+class SimulatedTimeAwarePlatformUtil : public PlatformUtil {
+public:
+  SimulatedTimeAwarePlatformUtil();
+  virtual ~SimulatedTimeAwarePlatformUtil();
```
htuch: Nit: doesn't need to be `virtual` as these are effectively final classes.

nighthawk/test/mocks.h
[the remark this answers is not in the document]
htuch: Yep, all good.

nighthawk/test/mocks.h
htuch: I think it would be best to separate the mock from the fake (simulated) implementation. You can still have individual test cases use the mock, and instantiate it to delegate as needed, it just feels a bit icky baking in one particular delegated behavior into the mock.

Mocks should generally be lightweight and policy free, although in Envoy we violate this quite a bit and do some lightweight simple behaviors; generally not fully delegation to a different class though.

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
oschaaf: Nice, `spin_timer_` makes sense, done. 

Regarding always spinning: I actually tried that a little while ago. When doing that we unfortunately won't get any I/O events. So getting that to work may need a change in libevent or its configuration when we set it up. I do remember code mentioning priorities in the libevent code base, and in this particular perhaps prioritizing I/O events over timers would be what we need.

nighthawk/source/common/sequencer_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+
+  if (!from_periodic_timer) {
+    if (targets_initiated_ == targets_completed_) {
+      // We saturated the rate limiter, and there's no outstanding work.
+      // That means it looks like we are idle. Spin this event to improve
+      // accuracy. As a side-effect, this may help prevent CPU frequency scaling
+      // due to c-state changes. But on the other hand it may cause thermal throttling.
+      // TODO(oschaaf): Ideally we would have much finer grained timers instead.
+      // TODO(oschaaf): Optionize performing this spin loop.
+      platform_util_.yieldCurrentThread();
```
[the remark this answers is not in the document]
oschaaf: Maybe that not yielding would be another nice option to have? It *does* behave a little annoying in the current tests, because we don't have a place to move simulated time forward when we drop the call, but that is solvable. 

On a machine tuned for running NH cores could be isolated for it to run on, in which case yielding is practically guaranteed to be a no-op, as there should be no other thread to yield to. [1]

Through some quick manual testing, I could not find an obvious difference in measured timings.I think this probably does make a difference in some setups/scenarios though (noisy systems with lots of things going on? single core/cpu?).

[1] http://man7.org/linux/man-pages/man2/sched_yield.2.html
```
       If the calling thread is the only thread in the highest priority list
       at that time, it will continue to run after a call to sched_yield().
```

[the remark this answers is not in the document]
htuch: I think this is fine, I was just curious, nothing actionable beyond what you wrote above, thanks.

nighthawk/source/common/sequencer_impl.cc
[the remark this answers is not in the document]
htuch: That's interesting. Do we lose I/O events when spinning for rate limiter admittance as well? In any case, I'm sold on what you have here as a starting point, what I think we need to do to get this merged is to transfer this deep dive that we've been going through into a bunch of comments in the `.h` and `.cc`; most readers won't have any of this intuition! :)

nighthawk/test/sequencer_test.cc
```diff
@@ -0,0 +1,228 @@
...
+                          sequencer_target_, 1s, 1s);
+  EXPECT_CALL(getRateLimiter(), tryAcquireOne());
+  sequencer.start();
+  ASSERT_DEATH(sequencer.start(), "");
+}
+
+TEST_F(SequencerTest, WaitButNotStartedAsserts) {
+  SequencerImpl sequencer(platform_util_, *dispatcher_, time_system_, getRateLimiter(),
+                          sequencer_target_, 1s, 1s);
+  ASSERT_DEATH(sequencer.waitForCompletion(), "");
```
htuch: Validating `ASSERT` behavior is a bit strange, since you're driving the system into behavior that is explicitly never expected to happen. I think you can safely skip these.

nighthawk/test/sequencer_test.cc
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  sequencer.waitForCompletion();
+
+  // The test is actually that we get here. We would hang if the timeout didn't work. In any case,
+  // the test itself should have seen all callbacks...
+  EXPECT_EQ(5, callback_test_count_);
+  // ... but they ought to have not arrived at the Sequencer.
+  EXPECT_EQ(0, sequencer.latencyStatistic().count());
+  EXPECT_EQ(0, sequencer.blockedStatistic().count());
```
htuch: Do these tests cover the grace timeout, etc?

nighthawk/test/sequencer_test.cc
```diff
@@ -0,0 +1,228 @@
...
+  for (uint64_t i = 0; i < test_number_of_intervals_; i++) {
+    moveClockForwardOneInterval();
+  }
+
+  sequencer.waitForCompletion();
+  EXPECT_EQ(0, sequencer.latencyStatistic().count());
+  EXPECT_EQ(1, sequencer.blockedStatistic().count());
+}
+
+TEST_F(SequencerIntegrationTest, TimeoutTest) {
```
htuch: Can you add short `//` descriptions above each `TEST_F`?

nighthawk/test/mocks.h
htuch: I'm willing to drop this one for now, looking at the code again, I feel that this provides the required test capabilities and we can iterate later on the structure as we get more use cases for the mock time.

oschaaf:
@htuch updated this to address your last two comments. Nothing left on my end, ready to merge as far as I am concerned.  (ps. just also merged master into this branch)

PR #50, opened 2019-02-25
Statistics: fix a bug in `::toProto()` with empty Statistic, add test.

oschaaf:
Attempting to serialize a `Statistic` with zero values added would
result in a vague complaint: ""JsonObjectWriter was not fully closed".

Some sleuthing pintpointed this to `std::nan` values returned by
`mean()` and `pstdev()`, which would then be passed to
`set_nanos()` in proto duration fields.

### DIFF
--- a/nighthawk/source/common/statistic_impl.cc
+++ b/nighthawk/source/common/statistic_impl.cc
@@ -16,8 +16,8 @@ std::string StatisticImpl::toString() const {
 nighthawk::client::Statistic StatisticImpl::toProto() {
   nighthawk::client::Statistic statistic;
   statistic.set_count(count());
-  statistic.mutable_mean()->set_nanos(std::round(mean()));
-  statistic.mutable_pstdev()->set_nanos(std::round(pstdev()));
+  statistic.mutable_mean()->set_nanos(count() == 0 ? 0 : std::round(mean()));
+  statistic.mutable_pstdev()->set_nanos(count() == 0 ? 0 : std::round(pstdev()));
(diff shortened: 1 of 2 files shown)


nighthawk/source/common/statistic_impl.cc
```diff
@@ -16,8 +16,8 @@ std::string StatisticImpl::toString() const {
 nighthawk::client::Statistic StatisticImpl::toProto() {
   nighthawk::client::Statistic statistic;
   statistic.set_count(count());
-  statistic.mutable_mean()->set_nanos(std::round(mean()));
-  statistic.mutable_pstdev()->set_nanos(std::round(pstdev()));
+  statistic.mutable_mean()->set_nanos(count() == 0 ? 0 : std::round(mean()));
+  statistic.mutable_pstdev()->set_nanos(count() == 0 ? 0 : std::round(pstdev()));
```
htuch: Good catch. Can you meaningfully compute std dev if `count() == 1`?

oschaaf: Well, as I understand the result of `pstdev()` should be `0` when `count() == 1` holds. For `stdev()` it would be undefined/nan/etc.  We check for that here:

https://github.com/envoyproxy/envoy-perf/blob/453a895285603566f741a1ebf466e1d44f113438/nighthawk/test/statistic_test.cc#L90

Some discussion on the topic:
https://stackoverflow.com/questions/8023878/standard-deviation-of-one-element

PR #51, opened 2019-03-04
Nighthawk: sync up what we have right now with we-amp/nighthawk

oschaaf:
- Update Envoy submodule
- Fix hdrhistogram_c build breakage with later Envoy versions
- Add and use a few typedefs
- Update output.proto
- Tidy up some #includes
- Add an 'id' field to Statistic which helps use them in a more generic way + test
- Sequencer takes ownership of the RateLimiter that gets passed in

### DIFF
--- a/nighthawk/include/nighthawk/common/sequencer.h
+++ b/nighthawk/include/nighthawk/common/sequencer.h
@@ -1,3 +1,4 @@
+
 #pragma once
 
 #include <functional>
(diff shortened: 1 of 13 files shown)


nighthawk/include/nighthawk/common/statistic.h
```diff
@@ -63,9 +63,24 @@ class Statistic : Envoy::NonCopyable {
    * single global view. Types of the Statistics objects that will be combined
    * must be the same, or else a std::bad_cast exception will be raised.
    * @param statistic The Statistic that should be combined with this instance.
-   * @return std::unique_ptr<Statistic> instance.
+   * @return StatisticPtr instance.
    */
-  virtual std::unique_ptr<Statistic> combine(const Statistic& statistic) PURE;
+  virtual std::unique_ptr<Statistic> combine(const Statistic& statistic) const PURE;
```
htuch: Nit:pProbably could define the `StatisticPtr` above using forward decl of `Statistic`.

nighthawk/include/nighthawk/common/sequencer.h
```diff
@@ -31,16 +32,14 @@ class Sequencer {
...
-   * (In other words, time spend while the Sequencer is idle and not blocked by a rate limiter)
-   */
-  virtual const Statistic& blockedStatistic() const PURE;
-
-  /**
-   * @return const Statistic& tracks latency between calling the SequencerTarget and observing its
-   * callback.
+   * Gets a vector of associated Statistics.
+   *
+   * @return StatisticPtrVector A vector of Statistics.
```
htuch: Why is vector the natural representation? Why merge the two? Why not a map? I think it's because ultimately this is just one of a `repeated` in the output proto, but not sure if it makes sense to abandon typed structure there if there's only two of these. What's the future plans here?

oschaaf: Switched to a map, which I agree makes more sense. 

To generate cli/proto output, the current plan is that `Client` will pull the stats from one or more `Worker`(s). These in turn will pull stats from both associated `Sequencer` and `BenchmarkClient`.  The client will then merge stats into an aggregated view, next to the per-worker views.

oschaaf:
@htuch pushed a commit to address your comments, plus a fix for an issue where the Statistics proto output could overflow duration fields.

PR #53, opened 2019-03-10
StreamDecoder & MinimalTransportSocketFactoryContext

oschaaf:


nighthawk/source/client/stream_decoder.cc
```diff
@@ -0,0 +1,55 @@
...
+
+void StreamDecoder::decodeData(Envoy::Buffer::Instance&, bool end_stream) {
+  ASSERT(!complete_);
+  complete_ = end_stream;
+  if (complete_) {
+    onComplete(true);
+  }
+}
+
+void StreamDecoder::decodeTrailers(Envoy::Http::HeaderMapPtr&&) { NOT_IMPLEMENTED_GCOVR_EXCL_LINE; }
```
htuch: How come we never see trailers?

nighthawk/source/client/stream_decoder.cc
```diff
@@ -0,0 +1,55 @@
...
+void StreamDecoder::onResetStream(Envoy::Http::StreamResetReason) { onComplete(false); }
+
+void StreamDecoder::onPoolFailure(Envoy::Http::ConnectionPool::PoolFailureReason reason,
+                                  Envoy::Upstream::HostDescriptionConstSharedPtr) {
+  decoder_completion_callback_.onPoolFailure(reason);
+}
+
+void StreamDecoder::onPoolReady(Envoy::Http::StreamEncoder& encoder,
+                                Envoy::Upstream::HostDescriptionConstSharedPtr) {
+  encoder.encodeHeaders(request_headers_, true);
```
htuch: Might be good to add some comments here.. This is sending the request headers upstream? When does the rest of the request get sent?

nighthawk/source/client/stream_decoder.h
```diff
@@ -0,0 +1,80 @@
...
+
+  Envoy::TimeSource& time_source_;
+  StreamDecoderCompletionCallback& decoder_completion_callback_;
+  std::function<void()> caller_completion_callback_;
+  Statistic& connect_statistic_;
+  Statistic& latency_statistic_;
+  const Envoy::Http::HeaderMap& request_headers_;
+  Envoy::Http::HeaderMapPtr response_headers_;
+
+  Envoy::MonotonicTime connect_start_;
```
htuch: This can be `const` I think.

nighthawk/source/client/stream_decoder.cc
```diff
@@ -0,0 +1,55 @@
...
+void StreamDecoder::decodeTrailers(Envoy::Http::HeaderMapPtr&&) { NOT_IMPLEMENTED_GCOVR_EXCL_LINE; }
+
+void StreamDecoder::onComplete(bool success) {
+  if (success && measure_latencies_) {
+    latency_statistic_.addValue((time_source_.monotonicTime() - request_start_).count());
+  }
+  ASSERT(complete_);
+  decoder_completion_callback_.onComplete(success, *response_headers_);
+  caller_completion_callback_();
+  delete this;
```
htuch: FWIW, I've found historically self-deleting to be a dangerous pattern when combined with polymorphism, since any subclass needs to ensure it doesn't have any code after the `delete`. Did you consider using Envoy's `deferredDelete`? This is a much safer RCU-like approach. If you want to stick with self-deleting, I recommend making the class `final`.

nighthawk/test/options_test.cc
```diff
@@ -24,6 +26,14 @@ class OptionsImplTest : public testing::Test {
     return std::make_unique<OptionsImpl>(argv.size(), argv.data());
   }
 
+  // We suppress stderr because errors send by tclap mess up the google test output.
+  static void SetUpTestSuite() {
```
htuch: Is this for debug or permanent? What errors do we get from `tclap`?

nighthawk/source/client/stream_decoder.h
oschaaf: done

nighthawk/source/client/stream_decoder.cc
oschaaf: Nice, done

nighthawk/source/client/stream_decoder.cc
oschaaf: Fixed, plus I added some standalone testing for StreamDecoder in this PR.

nighthawk/source/client/stream_decoder.cc
oschaaf: So, currently this supports the simplest of cases only, sending GET requests. I was planning to wire through a to-be-designed `(Http)RequestSynthesizer` in the future to allow for more functionality.

nighthawk/test/options_test.cc
oschaaf: This was to work around a specific issue to vscode's terminal rendering getting confused. I removed it,  regular terminals don't have that problem. `tclap` reacts to bad input by sending help and complaints to stderr, which the tests do extensively.

nighthawk/source/client/stream_decoder.cc
htuch: Ack, maybe track it in the future work and we can resolve.

nighthawk/test/options_test.cc
htuch: Is there a way to redirect its logs to Envoy's? We do this in some places, e.g. with nghttp2, see https://github.com/envoyproxy/envoy/blob/1f9e2fd690e3747017f9e7aa8e6368592f5c71e7/source/common/http/http2/codec_impl.cc#L41

ci/do_ci.sh
```diff
@@ -34,20 +36,22 @@ if [ -n "$CIRCLECI" ]; then
...
+    CONCURRENCY=6
 else
     export PATH=/usr/lib/llvm-7/bin:$PATH
     export CC=clang
     export CXX=clang++
 fi
 
+export BAZEL_BUILD_OPTIONS="${BAZEL_BUILD_OPTIONS} --jobs ${CONCURRENCY}"
+export BAZEL_TEST_OPTIONS="${BAZEL_TEST_OPTIONS} --jobs ${CONCURRENCY} --local_test_jobs=${CONCURRENCY}"
+export MAKEFLAGS="-j ${CONCURRENCY}"
```
htuch: Unrelated to this PR, but how come we have `MAKEFLAGS` here?

nighthawk/source/client/stream_decoder.cc
```diff
@@ -0,0 +1,63 @@
...
+  onComplete(true);
+}
+
+void StreamDecoder::onComplete(bool success) {
+  if (success && measure_latencies_) {
+    latency_statistic_.addValue((time_source_.monotonicTime() - request_start_).count());
+  }
+  ASSERT(!success || complete_);
+  decoder_completion_callback_.onComplete(success, *response_headers_);
+  if (success) {
```
htuch: Should we be computing success/failure stats?

nighthawk/test/stream_decoder_test.cc
```diff
@@ -0,0 +1,136 @@
...
+  EXPECT_CALL(stream_encoder, encodeHeaders(HeaderMapEqualRef(&request_headers_), true));
+  decoder->onPoolReady(stream_encoder, ptr);
+  auto headers = std::make_unique<Envoy::Http::HeaderMapImpl>();
+  decoder->decodeHeaders(std::move(headers), true);
+  EXPECT_EQ(0, connect_statistic_.count());
+  EXPECT_EQ(0, latency_statistic_.count());
+}
+
+TEST_F(StreamDecoderTest, LatencyIsMeasured) {
+  auto decoder = new StreamDecoder(*dispatcher_, time_system_, *this, []() {}, connect_statistic_,
```
htuch: Who is going to be the ultimate owner of `StreamDecoder`? I'm curious if we can do a more robust ownership management here with smart pointers, generally we shy away from raw pointer in Envoy, but I agree that sometimes it makes sense. Just want to see if we can avoid potential leak scenarios.

oschaaf: I was considering adding a `StreamDecoderPool`; this may help avoid some allocations in the fast path, and would also be a natural place to take ownership. wdyt?

nighthawk/source/client/stream_decoder.cc
oschaaf: `BenchmarkClientHttpImpl` will do that for us:
https://github.com/envoyproxy/envoy-perf/pull/38/files#diff-28811197472bd6ba614ce0f99b28024fR210

Also, we leverage Envoy:

https://github.com/envoyproxy/envoy-perf/pull/38/files#diff-547dca47cd8b31225f4d73460ac4af1eR234

In the future, we could also consider plugging in something that aggregates counts per response status or group (4xx, 5xx). We could make that generic by having a function inspect the response headers, and returning a stat name that should be incremented?

nighthawk/source/client/stream_decoder.cc
oschaaf: Added to https://github.com/envoyproxy/envoy-perf/issues/32

ci/do_ci.sh
oschaaf: that was me killing the CI concurrency/OOM issue with more fire then needed I guess, that doesn't make sense. Removed.

nighthawk/test/options_test.cc
oschaaf: It looks like that doing that would be pretty straightforward:

1. Implement this interface: https://github.com/maddouri/tclap/blob/1e1cc4fb9abbc4bfcd62c73085c3c446fca681dd/include/tclap/CmdLineOutput.h
2. Use that implementation by calling `setOutput()` on the `TCLAP::CmdLine` instance.

oschaaf:
@htuch Code-level comments addressed, plus I left some feedback in comments.

nighthawk/test/stream_decoder_test.cc
htuch: Sure, that could be something to consider as an optimization. I'm curious about what to do even before we have this, is it possible that the allocating entity can also do cleanup via a smartptr, or do we need to self-destruct?

nighthawk/source/client/stream_decoder.cc
htuch: SG

nighthawk/source/client/stream_decoder.cc
```diff
@@ -0,0 +1,68 @@
...
+#include "common/http/http1/codec_impl.h"
+#include "common/http/utility.h"
+
+namespace Nighthawk {
+namespace Client {
+
+void StreamDecoder::decodeHeaders(Envoy::Http::HeaderMapPtr&& headers, bool end_stream) {
+  ASSERT(!complete_);
+  complete_ = end_stream;
+  response_headers_ = std::move(headers);
```
htuch: FWIW, you will also need to track `response_trailers_` and look at `grpc-status` when doing gRPC load tests later on. It's fine not to do this for now, maybe just leave a TODO.

PR #54, opened 2019-03-15
BenchmarkClient & WorkerImpl

oschaaf:
This PR contains the benchmark client and threader worker


nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,230 @@
...
+           .canCreate() ||
+      // In closed loop mode we want to be able to control the pacing as exactly as possible.
+      // In open-loop mode we probably want to skip this.
+      // NOTE(oschaaf): We can't consistently rely on resourceManager()::requests() because that
+      // isn't used for h/1 (it is used in tcp and h2 though).
+      ((requests_initiated_ - requests_completed_) >= connection_limit_)) {
+    return false;
+  }
+
+  auto stream_decoder = new StreamDecoder(
```
oschaaf: @htuch re: "is it possible that the allocating entity can also do cleanup via a smartptr, or do we need to self-destruct"

I'm not sure, as I understand it `StreamDecoder` relies on Envoy to perform the right callbacks, which will at some point drive it into its terminal state (and `StreamDecoder` is responsible for correctly interpreting these callbacks and determine when it has served its purpose). 

I wonder, wouldn't it be nice if somehow we could optionally transfer or share ownership of the decoder to/with Envoy in the call to `newStream` below? Then we could choose here: we transfer ownership (no pooling), or we share ownership and re-use it  (e.g. recycle to a decoder pool) once Envoy is done with it.

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,230 @@
...
+    Envoy::ProtobufTypes::MessagePtr message =
+        Envoy::Config::Utility::translateToFactoryConfig(transport_socket, config_factory);
+
+    ssl_context_manager_.reset(
+        new Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl(api_.timeSource()));
+    transport_socket_factory_context_ = std::make_unique<Ssl::MinimalTransportSocketFactoryContext>(
+        store_->createScope("client."), dispatcher_, generator_, *store_, api_,
+        *ssl_context_manager_);
+
+    // TODO(oschaaf): We perform some bootstrapping ourselves here, to avoid an assert during the
```
htuch: What do you mean by bootstrapping?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,230 @@
...
+    // integration test because of a runtime/tls conflict with the integration test server, when
+    // the message gets validated. Ideally we'd just re-use Tls::Upstream::createTransportFactory()
+
```
htuch: Nit: superfluous space.

nighthawk/source/client/benchmark_client_impl.h
```diff
@@ -0,0 +1,101 @@
...
+  std::unique_ptr<Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl>
+      ssl_context_manager_;
+  Envoy::Upstream::ClusterInfoConstSharedPtr cluster_;
+  StatisticPtr connect_statistic_;
+  StatisticPtr response_statistic_;
+  const bool use_h2_;
+  const std::unique_ptr<Uri> uri_;
+  Envoy::Network::Address::InstanceConstSharedPtr target_address_;
+  std::chrono::seconds timeout_;
+  uint64_t connection_limit_;
```
htuch: Tiny nit: I find it easier to verify we didn't skip initializing any of these if you do something like `uint64_t stream_reset_count_{};` to initialize here rather than constructor. For `const`, any non-trivial initializer that depends on args or non-POD, it's fine to do it in constructor.

nighthawk/source/common/worker_impl.h
```diff
@@ -0,0 +1,39 @@
...
+  void start() override;
+  void waitForCompletion() override;
+
+protected:
+  Envoy::Thread::ThreadFactory& thread_factory_;
+  Envoy::Event::DispatcherPtr dispatcher_;
+  Envoy::ThreadLocal::Instance& tls_;
+  Envoy::Stats::StorePtr store_;
+  std::unique_ptr<Envoy::Runtime::Loader> runtime_;
+  std::unique_ptr<Envoy::Runtime::RandomGenerator> generator_;
```
htuch: Should these just be owned objects? Also, Envoy has a singleton runtime now, should we just match that model?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,230 @@
...
+          const std::list<Envoy::Network::Address::InstanceConstSharedPtr>&& address_list) -> void {
+        active_dns_query_ = nullptr;
+        ENVOY_LOG(debug, "DNS resolution complete for {} ({} entries).", uri_->host_without_port(),
+                  address_list.size());
+        if (!address_list.empty()) {
+          target_address_ =
+              Envoy::Network::Utility::getAddressWithPort(*address_list.front(), uri_->port());
+          dns_resolved = true;
+        } else {
+          ENVOY_LOG(critical, "Could not resolve host [{}]", uri_->host_without_port());
```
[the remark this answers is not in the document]
oschaaf: Yes this is tested https://github.com/envoyproxy/envoy-perf/pull/54/files#diff-547dca47cd8b31225f4d73460ac4af1eR280

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,230 @@
...
+                                                 const std::string& uri, bool use_h2)
+    : api_(api), dispatcher_(dispatcher), store_(std::move(store)),
+      connect_statistic_(std::move(connect_statistic)),
+      response_statistic_(std::move(response_statistic)), use_h2_(use_h2),
+      uri_(std::make_unique<Uri>(Uri::Parse(uri))), timeout_(5s), connection_limit_(1),
+      max_pending_requests_(1), pool_overflow_failures_(0), stream_reset_count_(0),
+      requests_completed_(0), requests_initiated_(0), measure_latencies_(false) {
+  ASSERT(uri_->isValid());
+
+  connect_statistic_->setId("benchmark_http_client.queue_to_connect");
```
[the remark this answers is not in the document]
oschaaf: I think it makes sense to take a look at just using Envoy's statistics functionality, and figure out what it takes to switch. One thing I can think of is that we would need to add support for HdrHistogram to Envoy.

nighthawk/source/common/worker_impl.h
oschaaf: yes, done, now using the singleton runtime.

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: Updated this comment about bootstrapping

nighthawk/include/nighthawk/client/benchmark_client.h
```diff
@@ -0,0 +1,76 @@
...
+  virtual bool tryStartOne(std::function<void()> caller_completion_callback) PURE;
+
+  /**
+   * Transforms statistics matching the filter argument into a string of statistic "name:value"
+   * pairs, one per line.
+   * @param filter function that returns true iff a statistic should be transformed, based on the
+   * named and value it gets passed.
+   * @return std::string containing zero or more lines containing "name:value\n".
+   */
+  virtual std::string countersToString(CounterFilter filter) const PURE;
```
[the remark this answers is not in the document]
htuch: @oschaaf did we resolve to fix this in this PR or file an issue and do this later?

nighthawk/include/nighthawk/client/benchmark_client.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Determines if latency measurement is on.
+   *
+   * @return true if latency measurement is on.
```
htuch: Nit: you can condense this entire comment to:
```
/**
 * @return bool indicating if latency measurement is enabled.
 */
```

nighthawk/include/nighthawk/common/worker.h
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Wait for the worker thread to complete its work.
+   */
+  virtual void waitForCompletion() PURE;
+
+  /**
+   * Perform the actual work on the associated thread initiated by start().
+   */
+  virtual void work() PURE;
```
htuch: Why is this in the interface? Isn't it only a `WorkerImpl` detail?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,226 @@
+#include "nighthawk/source/client/benchmark_client_impl.h"
+
+#include "absl/strings/str_split.h"
+
+#include "ares.h"
```
htuch: How come we need Ares directly here (vs. via Envoy's DNS libs)?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,226 @@
...
+    } else {
+      common_tls_context->add_alpn_protocols("http/1.1");
+    }
+    auto transport_socket = cluster_config.transport_socket();
+    if (!cluster_config.has_transport_socket()) {
+      ASSERT(cluster_config.has_tls_context());
+      transport_socket.set_name(
+          Envoy::Extensions::TransportSockets::TransportSocketNames::get().Tls);
+      Envoy::MessageUtil::jsonConvert(cluster_config.tls_context(),
+                                      *transport_socket.mutable_config());
```
htuch: Nit: the more modern way to do this in Envoy is with typed config (`Any`), so that you don't need to round-trip with JSON.

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,226 @@
...
+        dispatcher_, host, Envoy::Upstream::ResourcePriority::Default, options);
+  }
+  return true;
+}
+
+void BenchmarkClientHttpImpl::terminate() { pool_.reset(); }
+
+StatisticPtrMap BenchmarkClientHttpImpl::statistics() const {
+  StatisticPtrMap statistics;
+  statistics[connect_statistic_->id()] = connect_statistic_.get();
```
htuch: Just a placeholder comment re: our discussion on filing an issue on better integration with Envoy stats.

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,226 @@
...
+void BenchmarkClientHttpImpl::onComplete(bool success, const Envoy::Http::HeaderMap& headers) {
+  requests_completed_++;
+  if (!success) {
+    // TODO(oschaaf): this information must bubble up!
+    stream_reset_count_++;
+  } else {
+    ASSERT(headers.Status());
+    const int64_t status = Envoy::Http::Utility::getResponseStatus(headers);
+    if (status >= 400 && status <= 599) {
+      // TODO(oschaaf): Figure out why this isn't incremented for us.
```
htuch: Happy to do some interactive debug on Slack on this one.

nighthawk/source/client/benchmark_client_impl.h
```diff
@@ -0,0 +1,103 @@
...
+class BenchmarkClientHttpImpl : public BenchmarkClient,
+                                public StreamDecoderCompletionCallback,
+                                public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
+public:
+  BenchmarkClientHttpImpl(Envoy::Api::Api& api, Envoy::Event::Dispatcher& dispatcher,
+                          Envoy::Stats::StorePtr&& store, StatisticPtr&& connect_statistic,
+                          StatisticPtr&& response_statistic, const std::string& uri, bool use_h2);
+
+  // TODO(oschaaf): can probably get rid of these.
+  uint64_t stream_reset_count() { return stream_reset_count_; }
```
htuch: Nit: this can be a `const` method`

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+#include "test/integration/integration.h"
+#include "test/integration/utility.h"
+#include "test/server/utility.h"
+#include "test/test_common/utility.h"
+
+using namespace std::chrono_literals;
+
+namespace Nighthawk {
+
+std::string lorem_ipsum_config;
```
htuch: Nit: this is technically a static non-PoD, which can be problematic due to the static initializaiton fiasco in C++. Also, prefer to avoid globals. Can this be inlined into `SetUpTestCase`?

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+                            public testing::TestWithParam<Envoy::Network::Address::IpVersion> {
+public:
+  BenchmarkClientTest()
+      : Envoy::BaseIntegrationTest(GetParam(), realTime(), lorem_ipsum_config),
+        api_(thread_factory_, store_, timeSystem()), dispatcher_(api_.allocateDispatcher()) {}
+
+  static void SetUpTestCase() {
+    Envoy::Filesystem::InstanceImpl filesystem;
+
+    Envoy::TestEnvironment::setEnvVar("TEST_TMPDIR", Envoy::TestEnvironment::temporaryDirectory(),
```
htuch: How come we need to do this? Doesn't Bazel set this?

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+
+    Envoy::TestEnvironment::exec({Envoy::TestEnvironment::runfilesPath("nighthawk/test/certs.sh")});
+
+    lorem_ipsum_config = filesystem.fileReadToEnd(Envoy::TestEnvironment::runfilesPath(
+        "nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml"));
+    lorem_ipsum_config = Envoy::TestEnvironment::substitute(lorem_ipsum_config);
+  }
+
+  void SetUp() override {
+    ares_library_init(ARES_LIB_INIT_ALL);
```
htuch: Consider factoring out this boiler plate in case other tests need it.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+  uint32_t getTestServerHostAndPort() { return lookupPort("listener_0"); }
+
+  uint32_t getTestServerHostAndSslPort() { return lookupPort("listener_1"); }
+
+  void TearDown() override {
+    client_->terminate();
+    test_server_.reset();
+    fake_upstreams_.clear();
+    tls_.shutdownGlobalThreading();
+    ares_library_cleanup();
```
htuch: Ditto.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+  void setupBenchmarkClient(const std::string uriPath, bool use_https, bool use_h2,
+                            uint32_t port = 0) {
+    if (port == 0) {
+      port = use_https ? getTestServerHostAndSslPort() : getTestServerHostAndPort();
+    }
+
+    client_ = std::make_unique<Client::BenchmarkClientHttpImpl>(
+        api_, *dispatcher_, std::make_unique<Envoy::Stats::IsolatedStoreImpl>(),
+        std::make_unique<StreamingStatistic>(), std::make_unique<StreamingStatistic>(),
+        fmt::format("{}://127.0.0.1:{}{}", use_https ? "https" : "http", port, uriPath), use_h2);
```
htuch: Nit: this will break in IPv6-only environments such as Google. This will become an issue during the Google import. See how this is handled safely with `TEST_P` and IpVersions in Envoy's `test/integration` tests.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+    Client::CounterFilter filter = [](std::string, uint64_t value) { return value > 0; };
+    return client_->countersToString(filter);
+  }
+
+  Envoy::Thread::ThreadFactoryImplPosix thread_factory_;
+  Envoy::Stats::IsolatedStoreImpl store_;
+  Envoy::Api::Impl api_;
+  Envoy::Event::DispatcherPtr dispatcher_;
+  Envoy::Runtime::RandomGeneratorImpl generator_;
+  ::testing::NiceMock<Envoy::ThreadLocal::MockInstance> tls_;
```
htuch: Nit: probably worth a `using ::testing::NiceMock` at the top of file.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+client.upstream_cx_tx_bytes_total:82\n\
+client.upstream_rq_pending_total:1\n\
+client.upstream_rq_total:1",
+            getNonZeroValuedCounters());
+}
+
+TEST_P(BenchmarkClientTest, BasicTestHttpsH1) {
+  testBasicFunctionality("/lorem-ipsum-status-200", 1, 1, true, false, 10);
+  EXPECT_EQ("client.ssl.ciphers.ECDHE-RSA-AES128-GCM-SHA256:1\n\
+client.ssl.curves.X25519:1\n\
```
htuch: Yeah, I reckon if you get back a proto here, it would be super easy to write individual EXPECTs or do a proto compare.

nighthawk/test/certs.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
htuch: Can you reuse the certs that already exist in https://github.com/envoyproxy/envoy/tree/master/test/config/integration/certs?

nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml
```diff
@@ -0,0 +1,89 @@
+admin:
+  access_log_path: /dev/null
+  address:
+    socket_address:
+      address: 127.0.0.1
```
htuch: Yeah, I reckon the IPv6 issue will also arise here.

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,230 @@
...
+  thresholds->mutable_max_pending_requests()->set_value(max_pending_requests_);
+
+  Envoy::Network::TransportSocketFactoryPtr socket_factory;
+
+  if (uri_->scheme() == "https") {
+    auto common_tls_context = cluster_config.mutable_tls_context()->mutable_common_tls_context();
+    if (use_h2_) {
+      common_tls_context->add_alpn_protocols("h2");
+    }
+    common_tls_context->add_alpn_protocols("http/1.1");
```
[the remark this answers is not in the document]
oschaaf: Yes, good catch, that was actually the way this worked before switching over to the new `ssl.h` we have now.

nighthawk/test/benchmark_http_client_test.cc
oschaaf: Sure, I'd love to, I thought that `using` should be avoided according to the style guide, but I'm not sure where I picked that up. I propose we do that in a separate PR and to enforce consistency accross the board. Added to the track-for-future list.

nighthawk/test/benchmark_http_client_test.cc
oschaaf: I switched this to a std::map, I'm planning to consolidate the translation to protoc into a single layer. SG?

nighthawk/test/benchmark_http_client_test.cc
oschaaf: This is fixed, I think, but I'm looking for a way to ensure/check we used `ipv4` / `ipv6` in test expectations.

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: Resolving this as per out-of-band discussion

nighthawk/test/certs.sh
oschaaf: Yes, indeed. I symlinked the certs we need, and after tweaking the environment variables, I was able to drop the script that generates them and use those instead.

nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml
oschaaf: I switched this to `{{ ip_loopback_address }}`, I think that is the the right thing to do here?

nighthawk/include/nighthawk/common/worker.h
oschaaf: Pinged you on slack to discuss

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: Added to the track-for-future issue

nighthawk/test/benchmark_http_client_test.cc
oschaaf: No, it's not added for us by Bazel, but I removed this here, and added it to the CI script instead, I think that is what Envoy does.

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: After looking at the code, it looks like this needs to be changed, the stat that is used is incremented upon actual protocol violations, not when the protocol indicates an error.

oschaaf: Added NH's own set of statistics to track counts for 1xx,2xx,..,5xx, resolving this.

nighthawk/test/benchmark_http_client_test.cc
oschaaf: Added this to the tracking issue for now

ci/do_ci.sh
```diff
@@ -36,6 +36,10 @@ if [ -n "$CIRCLECI" ]; then
         mv "${HOME:-/root}/.gitconfig" "${HOME:-/root}/.gitconfig_save"
         echo 1
     fi
+    export TEST_TMPDIR=/build/tmp
+    export TEST_CERT_DIR="$TEST_TMPDIR"
```
htuch: Is `TEST_CERT_DIR` used anywhere?

nighthawk/include/nighthawk/client/benchmark_client.h
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @return true if the request could be started.
+   * @return false if the request could not be started, for example due to resource limits.
+   */
+  virtual bool tryStartOne(std::function<void()> caller_completion_callback) PURE;
+
+  /**
+   * Gets a map of tracked counter values, keyed by name.
+   * @param filter function that returns true iff a counter should be included in the map,
+   * based on the named and value it gets passed. The default filter returns all counters.
```
htuch: FWIW, Envoy also does stats filtering with StatsMatcher.

htuch: https://github.com/envoyproxy/envoy/blob/4a5f85809cef13be47873483789a5a7bb64ded7c/api/envoy/config/metrics/v2/stats.proto#L76

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,243 @@
...
+                                            ? Envoy::Http::Headers::get().SchemeValues.Https
+                                            : Envoy::Http::Headers::get().SchemeValues.Http);
+}
+
+bool BenchmarkClientHttpImpl::syncResolveDns() {
+  bool dns_resolved = false;
+  auto dns_resolver = dispatcher_.createDnsResolver({});
+  std::string to_resolve = uri_->host_without_port();
+
+  if (to_resolve.size() > 0 && to_resolve[0] == '[') {
```
htuch: What is this for? Please add comment in source as well as review.

nighthawk/include/nighthawk/client/benchmark_client.h
htuch: Actually, maybe I misunderstand. Why do we need to filter here? Could we just return the raw stats object and use that?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,243 @@
...
+
+namespace Nighthawk {
+namespace Client {
+
+BenchmarkClientHttpImpl::BenchmarkClientHttpImpl(Envoy::Api::Api& api,
+                                                 Envoy::Event::Dispatcher& dispatcher,
+                                                 Envoy::Stats::StorePtr&& store,
+                                                 StatisticPtr&& connect_statistic,
+                                                 StatisticPtr&& response_statistic,
+                                                 const std::string& uri, bool use_h2)
```
htuch: Tiny nit: generally prefer `absl::string_view` over `const std::string&` in new code, since it's strictly more flexibly and compatible with the C++17 brace new world.

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,243 @@
...
+    } else if (status > 499 && status <= 599) {
+      benchmark_client_stats_.http_5xx_.inc();
+    } else {
+      benchmark_client_stats_.http_xxx_.inc();
+    }
+  }
+}
+
+void BenchmarkClientHttpImpl::onPoolFailure(Envoy::Http::ConnectionPool::PoolFailureReason reason) {
+  ASSERT(reason == Envoy::Http::ConnectionPool::PoolFailureReason::ConnectionFailure);
```
htuch: Is this more of a `NOT_IMPLEMENTED_UNLESS` or an actual invariant?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,243 @@
...
+    } else if (status > 199 && status <= 299) {
+      benchmark_client_stats_.http_2xx_.inc();
+    } else if (status > 299 && status <= 399) {
+      benchmark_client_stats_.http_3xx_.inc();
+    } else if (status > 399 && status <= 499) {
+      benchmark_client_stats_.http_4xx_.inc();
+    } else if (status > 499 && status <= 599) {
+      benchmark_client_stats_.http_5xx_.inc();
+    } else {
+      benchmark_client_stats_.http_xxx_.inc();
```
htuch: Envoy also tracks this, but I think it might be in HCM, which you won't have.

nighthawk/source/client/output.proto
```diff
@@ -26,4 +24,16 @@ message Statistic {
...
   google.protobuf.Duration pstdev = 4;
   repeated Percentile percentiles = 5;
-}
\ No newline at end of file
+}
+
+message Result {
+  string name = 1;
+  repeated Statistic statistics = 2;
+  repeated Counter counters = 3;
```
htuch: Envoy has stats dumping protos, see https://github.com/envoyproxy/envoy/blob/master/api/envoy/admin/v2alpha/metrics.proto, should we reference these instead?

nighthawk/source/common/utility.cc
```diff
@@ -34,21 +34,35 @@ Uri::Uri(const std::string& uri) : scheme_("http") {
...
   const size_t scheme_end = uri.find("://", 0);
-
   if (scheme_end != std::string::npos) {
     scheme_ = absl::AsciiStrToLower(uri.substr(0, scheme_end));
   }
 
+  size_t colon_index = std::string::npos;
+  bool in_ipv6_address = false;
+
+  for (size_t i = 0; i < host_and_port_.size(); i++) {
```
htuch: Don't any of the existing Envoy address parsing utils help here? I'm surprised we're the first to be deconstructing an IPv6 address..

htuch: If we need this, I suggest splitting out into a separate utility and adding tests..

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,336 @@
...
+        filesystem.fileReadToEnd(Envoy::TestEnvironment::runfilesPath(path));
+    Envoy::TestEnvironment::writeStringToFileForTest(path_to, content);
+  }
+
+  static void SetUpTestCase() {
+    ASSERT_NE("", Envoy::TestEnvironment::getCheckedEnvVar("TEST_TMPDIR"));
+
+    copyFileToWorkingDir("nighthawk/test/test_data/lorem_ipsum.txt", "lorem_ipsum.txt");
+    copyFileToWorkingDir("nighthawk/test/test_data/certs/cacert.pem", "cacert.pem");
+    copyFileToWorkingDir("nighthawk/test/test_data/certs/servercert.pem", "servercert.pem");
```
htuch: Why can't we reference these directly in the Envoy tree where they already exist via test data? That way, if we need to bump them, for example when they expire or we want to change some baseline assumptions about server certs, it's less churn.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,336 @@
...
+
+  static void copyFileToWorkingDir(const std::string path, const std::string path_to) {
+    Envoy::Filesystem::InstanceImpl filesystem;
+    const std::string content =
+        filesystem.fileReadToEnd(Envoy::TestEnvironment::runfilesPath(path));
+    Envoy::TestEnvironment::writeStringToFileForTest(path_to, content);
+  }
+
+  static void SetUpTestCase() {
+    ASSERT_NE("", Envoy::TestEnvironment::getCheckedEnvVar("TEST_TMPDIR"));
```
htuch: I think you can assume this from Bazel, since it sets this variable, https://docs.bazel.build/versions/master/test-encyclopedia.html

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,336 @@
...
+    ares_library_cleanup();
+  }
+
+  void setupBenchmarkClient(const std::string uriPath, bool use_https, bool use_h2,
+                            uint32_t port = 0) {
+    if (port == 0) {
+      port = use_https ? getTestServerHostAndSslPort() : getTestServerHostAndPort();
+    }
+
+    std::string address =
```
htuch: Nit: `const`

nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml
```diff
@@ -0,0 +1,89 @@
+admin:
```
htuch: Did you consider using https://github.com/envoyproxy/envoy/blob/master/test/config/utility.h to build this bootstrap? We try to generally avoid hardcoded bootstrap YAML in tests, but if you need gymnastics to make `ConfigHelper` work, this is probably OK.

nighthawk/test/benchmark_http_client_test.cc
oschaaf: got rid of these, let me know if this looks good now

nighthawk/source/common/utility.cc
oschaaf: Well, there's `Utility::parseInternetAddressAndPort()` (which we now use in the client later on). But it expects a port, which we do not always have (the code is working to figure that out here). Splitted this out and added tests; also added some limited tests for hostname validation

nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml
oschaaf: Tbh, I had not seen that, it looks useful. I rewrote some of the bootstrap which allowed a reduction of the amount of yaml needed.. Shall we add improving this further to the track-for-future issue?

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: This was checking to see if the hostname represented an ipv6 address; this code changed a bit and now avoids forwarding ip adresses to dns.

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: Yeah, added a tracking issue to replace the remaining `const std::string&` instances with `absl::string_view` in one go; resolving this for now.

nighthawk/source/client/benchmark_client_impl.cc
oschaaf: We have a bug when this asserts, in the current closed-loop mode we avoid pool overflows as we throttle ourselves. When open-loop testing lands, this needs to be revisited.

nighthawk/source/client/output.proto
oschaaf: Yes. I looked into that but I'm having trouble with the proto import. Adding this to the list to unblock here.

nighthawk/include/nighthawk/client/benchmark_client.h
oschaaf: We use the filter to get non-0 valued stats. We can use that to keep the output small, but also in tests to ensure that we check all relevant counters (when we want to).

nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml
htuch: Yep, sounds like a plane.

nighthawk/include/nighthawk/client/benchmark_client.h
```diff
@@ -0,0 +1,73 @@
...
+   * Terminate will be called on the worker thread before it ends.
+   */
+  virtual void terminate() PURE;
+
+  /**
+   * Turns latency measurement on or off.
+   *
+   * @param measure_latencies true iff latencies should be measured.
+   */
+  virtual void setMeasureLatencies(bool measure_latencies) PURE;
```
htuch: I wonder if this is exposing an internal implementation detail; why not always measure latencies?

nighthawk/include/nighthawk/client/benchmark_client.h
```diff
@@ -0,0 +1,73 @@
...
+   */
+  virtual StatisticPtrMap statistics() const PURE;
+
+  /**
+   * Tries to start a request. In open-loop mode this MUST always return true.
+   *
+   * @param caller_completion_callback The callback the client must call back upon completion of a
+   * successfully started request.
+   *
+   * @return true if the request could be started.
```
htuch: Nit: prefer `@return bool true if .. otherwise ...`

nighthawk/include/nighthawk/client/benchmark_client.h
```diff
@@ -0,0 +1,73 @@
...
+
+typedef std::function<bool(const std::string, const uint64_t)> CounterFilter;
+
+class BenchmarkClient {
+public:
+  virtual ~BenchmarkClient() = default;
+
+  /**
+   * Initialize will be called on the worker thread after it has started.
+   * @param runtime to be used during initialization.
```
htuch: Nit: document `@return`. What would a fail here do?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -0,0 +1,247 @@
...
+
+bool BenchmarkClientHttpImpl::syncResolveDns() {
+  ASSERT(uri_->isValid());
+
+  // If the host can be parsed as an ipv4 or ipv6 address, we return that directly.
+  try {
+    target_address_ = Envoy::Network::Utility::parseInternetAddressAndPort(uri_->host_and_port());
+    return true;
+  } catch (EnvoyException) {
+  }
```
htuch: Maybe comment `// We get here when we have a hostname to resolve instead.` in the `catch` braces.

nighthawk/source/common/utility.cc
```diff
@@ -29,26 +29,73 @@ uint32_t determineCpuCoresWithAffinity() {
...
+        colon_index = i;
+        break;
+      } else if (c == '.') {
+        if (last_seen_dot) {
+          throw InvalidHostException("Invalid hostname");
+        }
+        last_seen_dot = true;
+        continue;
+      } else {
+        last_seen_dot = false;
```
htuch: I reckon we probably don't need to do any of this validation; it's the user's responsibility to provide something valid, I think we only care about the `:` separator. Otherwise, we get in the business of doing what other libraries (e.g. Ares) are already doing. This then simplifies quite a bit.

nighthawk/source/common/utility.cc
```diff
@@ -29,26 +29,73 @@ uint32_t determineCpuCoresWithAffinity() {
...
+  size_t colon_index = std::string::npos;
+  bool in_ipv6_address = false;
+  bool last_seen_dot = true;
+
+  // TODO(oschaaf): We need to revisit this and improve. This performs some
+  // limited hostname validation.
+  for (size_t i = 0; i < authority.size(); i++) {
+    char c = authority[i];
+    if (i == 0 && c == '[') {
+      in_ipv6_address = true;
```
htuch: I'm probably super dense here, so apologies if we've covered this before, but what's the reason we can't use `Utility::parseInternetAddressAndPort` again?

nighthawk/source/common/utility.cc
```diff
@@ -29,26 +29,73 @@ uint32_t determineCpuCoresWithAffinity() {
...
   if (scheme_end != std::string::npos) {
     scheme_ = absl::AsciiStrToLower(uri.substr(0, scheme_end));
   }
 
+  size_t colon_index;
+
+  try {
+    colon_index = findPortSeparatorInAuthority(host_and_port_);
+  } catch (InvalidHostException ex) {
+    host_error_ = true;
```
htuch: Why not throw an exception in the constructor? This is the idiomatic C++ way to fail a constructor.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,410 @@
...
+    dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
+
+    EXPECT_EQ(0, getCounter("benchmark.stream_resets"));
+  }
+
+  virtual void setupBenchmarkClient(const std::string uriPath, bool use_h2) = 0;
+
+  void doSetupBenchmarkClient(const std::string uriPath, bool use_https, bool use_h2) {
+    const std::string address =
+        GetParam() == Envoy::Network::Address::IpVersion::v4 ? "127.0.0.1" : "[::1]";
```
htuch: See Envoy's `getLoopbackAddressString` util.

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,410 @@
...
+  void testBasicFunctionality(const std::string uriPath, uint64_t max_pending,
+                              uint64_t connection_limit, bool use_h2, uint64_t amount_of_request) {
+    setupBenchmarkClient(uriPath, use_h2);
+
+    client_->setConnectionTimeout(10s);
+    client_->setMaxPendingRequests(max_pending);
+    client_->setConnectionLimit(connection_limit);
+    EXPECT_TRUE(client_->initialize(runtime_));
+
+    uint64_t amount = amount_of_request;
```
htuch: Nit: `const`

nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml
```diff
@@ -0,0 +1,49 @@
...
+              - name: service
+                domains:
+                - "*"
+                routes:
+                - match:
+                    prefix: /lorem-ipsum-status-200
+                  direct_response:
+                    status: 200
+                    body:
+                      filename: nighthawk/test/test_data/lorem_ipsum.txt
```
htuch: Interesting that this works. I think we probably should do a runfiles based substitution here though, since we will need this when we import to Google (the monorepo path will have some additional prefix that we need a hook to substitute).

nighthawk/test/benchmark_http_client_test.cc
```diff
@@ -0,0 +1,410 @@
...
+    };
+
+    for (uint64_t i = 0; i < amount; i++) {
+      if (client_->tryStartOne(f)) {
+        inflight_response_count++;
+      }
+    }
+
+    EXPECT_EQ(max_pending, inflight_response_count);
+
```
htuch: Nit: prefer not to have this much superfluous whitepace. Google style prefers compactness.

nighthawk/source/common/utility.cc
htuch: Hmm, this answers my earlier comment. Can't we just do a quick scan from RHS, check if there is a `:`, and if not, add a dummy port?

nighthawk/include/nighthawk/client/benchmark_client.h
oschaaf: This gets used later, we currently do not measure during the warmup phase, and enable latency when we transition out the it

htuch: Ack, ignore this comment then. Thanks for the explanation.

nighthawk/include/nighthawk/client/benchmark_client.h
oschaaf: This no longer has a return value

oschaaf:
@htuch this is ready for another pass. I moved all address resolution (parsing, dns) into the Uri class, and simplified it. This fixed another todo: the address resolution in the `nighthawk` branch now happens on the main thread and the resolved uri is passed on to the workers (instead of doing resolution on each worker thread). Also improved the tests for the name resolution process (wrt ipv4/ipv6).

nighthawk/source/client/output.proto
htuch: Fair enough; would like to see this in the final NH but fine to leave for now.

nighthawk/source/common/utility.cc
```diff
@@ -29,28 +29,97 @@ uint32_t determineCpuCoresWithAffinity() {
...
+    return hostname.find(":", hostname.find(']'));
+  }
+  return hostname.rfind(":");
+}
+
+Uri::Uri(absl::string_view uri) : scheme_("http") {
   absl::string_view host, path;
   Envoy::Http::Utility::extractHostPathFromUri(uri, host, path);
+
+  if (host.size() == 0) {
```
htuch: Nit: `host.empty()`

nighthawk/source/common/utility.cc
```diff
@@ -29,28 +29,97 @@ uint32_t determineCpuCoresWithAffinity() {
 
 } // namespace PlatformUtils
 
-Uri::Uri(const std::string& uri) : scheme_("http") {
+size_t Uri::findPortSeparator(absl::string_view hostname) {
```
htuch: +1 on this reorg to `Uri`, looks much cleaner now!

nighthawk/source/common/utility.cc
```diff
@@ -29,28 +29,97 @@ uint32_t determineCpuCoresWithAffinity() {
...
+  }
+  resolve_attempted_ = true;
+
+  // First we attempt to parse the hostname as host:port. Iff that didn't work out, we try to
+  // resolve using dns.
+  bool ok = tryParseHostAsAddressAndPort(dns_lookup_family) ||
+            performDnsLookup(dispatcher, dns_lookup_family);
+
+  // Ensure that we figured out a fitting match for the requested dns lookup family.
+  ok = ok && !((dns_lookup_family == Envoy::Network::DnsLookupFamily::V6Only &&
```
htuch: Is it possible to trigger these conditions, i.e. if we ask for IPv6 only at DNS resolution, might we actually get back an IPv4?

nighthawk/test/worker_test.cc
```diff
@@ -0,0 +1,61 @@
...
+#include "common/api/api_impl.h"
+#include "common/filesystem/filesystem_impl.h"
+#include "common/runtime/runtime_impl.h"
+#include "common/stats/isolated_store_impl.h"
+
+#include "test/mocks/thread_local/mocks.h"
+
+#include "nighthawk/test/mocks.h"
+
+#include "nighthawk/source/common/worker_impl.h"
```
htuch: We really need something like Envoy's check/fix_format scripts to enforce header ordering. Can you add this to the backlog if it doesn't already exist there?

nighthawk/test/utility_test.cc
```diff
@@ -41,17 +46,90 @@ TEST_F(UtilityTest, ExplicitPort) {
...
+  EXPECT_EQ("[::1]:81", u.host_and_port());
+  EXPECT_EQ(81, u.port());
+
+  const Uri u2 = Uri::Parse("http://[::1]/bar");
+  EXPECT_EQ("[::1]", u2.host_without_port());
+  EXPECT_EQ("[::1]:80", u2.host_and_port());
+  EXPECT_EQ(80, u2.port());
+}
+
+TEST_F(UtilityTest, findPortSeparator) {
```
htuch: Nit: `FindPortSeparator`

nighthawk/test/utility_test.cc
```diff
@@ -41,17 +46,90 @@ TEST_F(UtilityTest, ExplicitPort) {
...
+    auto u = Uri::Parse(uri);
+    return u.resolve(*dispatcher, address_family);
+  }
+};
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, UtilityAddressResolutionTest,
+                         testing::ValuesIn(Envoy::TestEnvironment::getIpVersionsForTest()),
+                         Envoy::TestUtility::ipTestParamsToString);
+
+TEST_P(UtilityAddressResolutionTest, addressResolution) {
```
htuch: Nit: can you fix up the capitalization on these other tests as well?

htuch:
@oschaaf P.S. can you lower the coverage threshold for now so that we can get the coverage HTML and verify the new code is fully tested?

nighthawk/source/common/utility.cc
oschaaf: The test case that attempts to resolve `[::1]:80` while requesting IPv4-only triggers it. But that doesn't go through DNS, and is handled by `Envoy::Network::Utility::parseInternetAddressAndPort()`

nighthawk/source/common/utility.cc
```diff
@@ -29,7 +29,7 @@ uint32_t determineCpuCoresWithAffinity() {
 
 } // namespace PlatformUtils
 
-size_t Uri::findPortSeparator(absl::string_view hostname) {
+size_t Uri::FindPortSeparator(absl::string_view hostname) {
```
htuch: I meant change the name in the test, not the implementation :P

htuch:
Somehow `run_coverage.sh` is conflicting, can you take a look?

nighthawk/source/common/utility.cc
```diff
@@ -29,28 +30,105 @@ uint32_t determineCpuCoresWithAffinity() {
 
 } // namespace PlatformUtils
 
-Uri::Uri(const std::string& uri) : scheme_("http") {
+Uri Uri::Parse(absl::string_view uri) {
```
htuch: FWIW, in a completely independent exercise, I was looking at Chromium's URL handling code, and it has all of this plus more. We're looking at how we can incorporate this Envoy-side long-term, so I feel that at some point, we can just switch to that and not have to maintain our own. Might be worth noting somewhere.

PR #66, opened 2019-03-28
Nighthawk: OutputFormatting

oschaaf:


nighthawk/source/client/output_formatter_impl.cc
```diff
@@ -0,0 +1,86 @@
...
+    : OutputFormatterImpl(time_source, options, merged_statistics, merged_counters) {}
+
+std::string ConsoleOutputFormatterImpl::toString() const {
+  std::stringstream s;
+
+  // TODO(oschaaf): echo non-default options to CLI output?
+  s << "Merged statistics:\n";
+  for (auto& statistic : merged_statistics_) {
+    if (statistic->count() > 0) {
+      s << fmt::format("{}: {}\n", statistic->id(), statistic->toString(), "\n");
```
htuch: Nit: why is the final "\n" an argument rather than part of the format string?

nighthawk/source/client/output_formatter_impl.cc
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+  s << "\nMerged counters\n";
+  for (auto counter : merged_counters_) {
+    s << fmt::format("counter {}:{}\n", counter.first, counter.second);
+  }
```
htuch: There's a part of me that wished there was some simple C++ string templating thing like Python's string templates. But, in the absence of that, this looks fairly readable.

nighthawk/source/client/output_formatter_impl.h
```diff
@@ -0,0 +1,54 @@
...
+namespace Nighthawk {
+namespace Client {
+
+class OutputFormatterImpl : public OutputFormatter {
+public:
+  OutputFormatterImpl(Envoy::TimeSource& time_source, const Options& options,
+                      const std::vector<StatisticPtr>& merged_statistics,
+                      const std::map<std::string, uint64_t>& merged_counters);
+
+  Envoy::ProtobufTypes::MessagePtr toProto() const;
```
htuch: This is public, but not part of the interface, can you elaborate on the reasoning?

nighthawk/source/client/output_formatter_impl.cc
```diff
@@ -0,0 +1,86 @@
...
+  return s.str();
+}
+
+Envoy::ProtobufTypes::MessagePtr OutputFormatterImpl::toProto() const {
+  nighthawk::client::Output output;
+  output.set_allocated_options(options_.toCommandLineOptions().release());
+
+  auto ts = time_source_.systemTime().time_since_epoch();
```
htuch: Nit: make all of these intermediates `const`

nighthawk/source/client/output_formatter_impl.cc
```diff
@@ -0,0 +1,86 @@
...
+  auto seconds = std::chrono::duration_cast<std::chrono::seconds>(ts);
+  output.mutable_timestamp()->set_seconds(seconds.count());
+  output.mutable_timestamp()->set_nanos(
```
htuch: There is some standard `Duration` and time conversion, see https://github.com/envoyproxy/envoy/blob/master/source/common/protobuf/utility.cc#L333 (you might need thereverse direction).

nighthawk/source/client/output_formatter_impl.cc
```diff
@@ -0,0 +1,86 @@
...
+
+JsonOutputFormatterImpl::JsonOutputFormatterImpl(
+    Envoy::TimeSource& time_source, const Options& options,
+    const std::vector<StatisticPtr>& merged_statistics,
+    const std::map<std::string, uint64_t>& merged_counters)
+    : OutputFormatterImpl(time_source, options, merged_statistics, merged_counters) {}
+
+std::string JsonOutputFormatterImpl::toString() const {
+  auto output = toProto();
+  return Envoy::MessageUtil::getJsonStringFromMessage(*output, true, true);
```
htuch: Nit: inline with above, same below.

nighthawk/source/client/output_formatter_impl.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <cstdint>
+
+namespace Nighthawk {
+namespace Client {
+
+class OutputFormatterImpl : public OutputFormatter {
+public:
+  OutputFormatterImpl(Envoy::TimeSource& time_source, const Options& options,
+                      const std::vector<StatisticPtr>& merged_statistics,
```
htuch: Can you add some comments on what "merged" means here for stats/counters?

htuch:
Also needs master merge.

nighthawk/source/client/output_formatter_impl.cc
oschaaf: Nice catch, that was a remnant of earlier stuff, and shouldn't be there at all. Removed.

nighthawk/source/client/output_formatter_impl.h
oschaaf: done

nighthawk/source/client/output_formatter_impl.cc
oschaaf: I think this is better now

nighthawk/source/client/output_formatter_impl.h
oschaaf: My bad, I changed it to be `protected`; there's been no need to expose it further at the moment. Lets defer that to if the need for doing so arises? (When that happens, we may want to relocate the `toProto` to it's own layer ?)

oschaaf:
@htuch merged master & comments addressed.

nighthawk/source/client/output_formatter_impl.cc
```diff
@@ -0,0 +1,79 @@
...
+  for (auto counter : merged_counters_) {
+    s << fmt::format("counter {}:{}\n", counter.first, counter.second);
+  }
+  return s.str();
+}
+
+Envoy::ProtobufTypes::MessagePtr OutputFormatterImpl::toProto() const {
+  nighthawk::client::Output output;
+  output.set_allocated_options(options_.toCommandLineOptions().release());
+  *(output.mutable_timestamp()) = google::protobuf::util::TimeUtil::NanosecondsToTimestamp(
```
htuch: This looks much nicer. Two things:

1. Can you use the `Envoy::Protobuf` namespace alias when referring to `google::protobuf` (here and elsewhere in NH if there are such places)? We need to do namespace remapping on Google import, and this assists with that. 

2. Where do you get that `time_since_epoch` is nanosecs here? Should we have an explicit duration cast?

oschaaf: 1. Done
2. Ugh; nice catch, added an explicit duration cast. On my system this resolves to nanoseconds..  but digging into this for a bit, this looks platform specific.

PR #67, opened 2019-03-29
Nighthawk: Add ClientWorkerXXX

oschaaf:
- Adds ClientWorker + tests
- Rate limiter improvement for supporting delayed starts
- WorkerImpl finalizes tls initialization by calling dispatcher.run

### DIFF
--- a/nighthawk/README.md
+++ b/nighthawk/README.md
@@ -20,7 +20,7 @@ First, follow steps 1 and 2 over at [Quick start Bazel build for developers](htt
 bazel build -c opt //nighthawk:nighthawk_client
 
 # test it
-bazel test -c opt //nighthawk/test:nighthawk_test
+bazel test -c fastbuild //nighthawk/test:nighthawk_test
 ```
 
 ## Using the Nighthawk client
@@ -162,6 +162,7 @@ The name of the file will be `<epoch.json>`, which contains:
 - The start time of the test, and a serialization of the Nighthawk options involved.
 - The mean latency and the observed standard deviation.
 - Latency percentiles produced by HdrHistogram.
+- Counters as tracked by Envoy's connection-pool and ssl libraries.
 
 ## Accuracy and repeatability considerations when using the Nighthawk client.
 
@@ -169,10 +170,10 @@ The name of the file will be `<epoch.json>`, which contains:
   processes that are not needed. 
 - Be aware that power state management and CPU Frequency changes are able to introduce significant noise.
   When idle, Nighthawk uses a busy loop to achieve precise timings when starting requests, which helps minimize this.
-  Still, consider disabling c-state changes in the system BIOS.
-- Be aware that CPU Thermal throttling may skew results.
-- Consider using `taskset` to isolate client and server. On machines with multiple physical CPU's there is a choice here.
-  You can partition client and server on the same physical processor, or run each of them on a different physical CPU.
+  Still, consider disabling C-state changes in the system BIOS.
+- Be aware that CPU thermal throttling may skew results.
+- Consider using `taskset` to isolate client and server. On machines with multiple physical CPUs there is a choice here.
+  You can partition client and server on the same physical processor, or run each of them on a different physical CPU. Be aware of the latency effects of interconnects such as QPI.
 - Consider disabling hyper-threading.
 - Consider tuning the benchmarking system for low latency (manually, or with tuned).
   - TODO(oschaaf): link resources.
--- a/nighthawk/README.md
+++ b/nighthawk/README.md
@@ -1,0 +1,183 @@
+# Nighthawk
+
+*A L7 HTTP protocol family benchmarking tool based on Envoy*
+
+## Current state
+
+The nighthawk client supports HTTP/1.1 and HTTP/2 over HTTP and HTTPS.
+
+HTTPS certificates are not yet validated
+
+## Prerequisites
+
+### Ubuntu
+
+First, follow steps 1 and 2 over at [Quick start Bazel build for developers](https://github.com/envoyproxy/envoy/blob/master/bazel/README.md#quick-start-bazel-build-for-developers).
+
+## Building and testing Nighthawk
+```bash
+# build it. for best accuracy it is important to specify -c opt.
+bazel build -c opt //nighthawk:nighthawk_client
+
+# test it
+bazel test -c opt //nighthawk/test:nighthawk_test
+```
+
+## Using the Nighthawk client
+
+```
+➜ bazel-bin/nighthawk_client --help
+
+USAGE:
+
+   bazel-bin/nighthawk_client  [-v <trace|debug|info|warn|error>]
+                               [--concurrency <string>] [--h2] [--timeout
+                               <uint64_t>] [--duration <uint64_t>]
+                               [--connections <uint64_t>] [--rps
+                               <uint64_t>] [--] [--version] [-h] <uri
+                               format>
+
+
+Where:
+
+   -v <trace|debug|info|warn|error|critical>,  --verbosity <trace|debug|info|warn
+      |error|critical>
+     Verbosity of the output. Possible values: [trace, debug, info, warn,
+     error, critical]. The default level is 'warn'.
+
+   --concurrency <string>
+     The number of concurrent event loops that should be used. Specify
+     'auto' to let Nighthawk leverage all vCPUs that have affinity to the Nighthawk
+     process. Note that increasing this results in an effective load multiplier
+     combined with the configured --rps and --connections values. Default: 1.
+
+   --h2
+     Use HTTP/2
+
+   --timeout <uint64_t>
+     Timeout period in seconds used for both connection timeout and grace
+     period waiting for lagging responses to come in after the test run is
+     done. Default: 5.
+
+   --duration <uint64_t>
+     The number of seconds that the test should run. Default: 5.
+
+   --connections <uint64_t>
+     The number of connections per event loop that the test should maximally use. Default:
+     1.
+
+   --rps <uint64_t>
+     The target requests-per-second rate. Default: 5.
+
+   --,  --ignore_rest
+     Ignores the rest of the labeled arguments following this flag.
+
+   --version
+     Displays version information and exits.
+
+   -h,  --help
+     Displays usage information and exits.
+
+   <uri format>
+     (required)  uri to benchmark. http:// and https:// are supported, but
+     in case of https no certificates are validated.
+
+
+   Nighthawk, a L7 HTTP protocol family benchmarking tool.
+```
+
+## Sample benchmark run
+
+```bash
+# start the benchmark target (Envoy in this case) on core 3.
+$ taskset -c 3 /path/to/envoy --config-path nighthawk/tools/envoy.yaml
+
+# run a quick benchmark using cpu cores 0 and 1.
+$ taskset -c 0-1 bazel-bin/nighthawk/nighthawk_client --connections 20 --rps 20000 --duration 5 --concurrency auto -v info http://127.0.0.1:10000/
+Nighthawk - A layer 7 protocol benchmarking tool.
+[23:11:09.078497][3431][I] [nighthawk/source/client/client.cc:69] Detected 2 (v)CPUs with affinity..
+[23:11:09.078521][3431][I] [nighthawk/source/client/client.cc:73] Starting 2 threads / event loops. Test duration: 5 seconds.
+[23:11:09.078524][3431][I] [nighthawk/source/client/client.cc:75] Global targets: 40 connections and 40000 calls per second.
+[23:11:09.078526][3431][I] [nighthawk/source/client/client.cc:79]    (Per-worker targets: 20 connections and 20000 calls per second)
+Merged statistics:
+benchmark_http_client.queue_to_connect: Count: 200205. Mean: 0.92 μs. pstdev: 0.79 μs.
+  Percentile          Value (usec)
+          50%          0.906
+          75%          0.928
+          90%          0.949
+          99%          1.009
+        99.9%          1.662
+       99.99%         48.649
+      99.999%         71.223
+         100%         72.327
+
+benchmark_http_client.request_to_response: Count: 200205. Mean: 35.35 μs. pstdev: 38.04 μs.
+  Percentile          Value (usec)
+          50%         32.745
+          75%         33.427
+          90%         34.077
+          99%         57.885
+        99.9%        514.143
+       99.99%        1526.65
+      99.999%        3703.81
+         100%        3737.86
+
+sequencer.blocking: Count: 149. Mean: 70.58 μs. pstdev: 68.95 μs.
+  Percentile          Value (usec)
+          50%         42.353
+          75%         92.499
+          90%        157.351
+          99%        309.935
+        99.9%        317.167
+       99.99%        317.167
+      99.999%        317.167
+         100%        317.167
+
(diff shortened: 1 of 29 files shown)


nighthawk/include/nighthawk/client/client_worker.h
```diff
@@ -0,0 +1,43 @@
...
+   */
+  virtual StatisticPtrMap statistics() const PURE;
+
+  /**
+   * @return const Envoy::Stats::Store& the statistics store associated the benchmark client.
+   */
+  virtual Envoy::Stats::Store& store() const PURE;
+
+  /**
+   * @brief Returns true iff the worker ran and completed successfully.
```
htuch: Nit: I think this can collapse to a single `@return`.

nighthawk/include/nighthawk/client/client_worker.h
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Gets the statistics, keyed by id.
+   * @return StatisticPtrMap A map of Statistics keyed by id.
+   */
+  virtual StatisticPtrMap statistics() const PURE;
+
+  /**
+   * @return const Envoy::Stats::Store& the statistics store associated the benchmark client.
+   */
+  virtual Envoy::Stats::Store& store() const PURE;
```
htuch: I guess a fundamental question is why we have to expose both a store and statistics.. I think this is because the `StatisticPtrMap` only handle very specialized statistics, rather than general "stats".

nighthawk/source/client/client_worker_impl.cc
```diff
@@ -0,0 +1,69 @@
...
+  // when comparing results to some other tools, which do open up the specified amount
+  // of connections.
+  benchmark_client_->tryStartOne([this] { dispatcher_->exit(); });
+  dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
+}
+
+void ClientWorkerImpl::delayStart() {
+  PlatformUtilImpl platform_util;
+  // TODO(oschaaf): We could use dispatcher to sleep, but currently it has a 1 ms resolution
+  // which is rather coarse for our purpose here.
```
htuch: How come this is too coarse? I would have though the initial delay would be O(seconds), so that would be fine. Would make the code a bit cleaner.

nighthawk/test/client_worker_test.cc
```diff
@@ -0,0 +1,114 @@
...
+  MockSequencer* sequencer_;
+  Envoy::Runtime::RandomGeneratorImpl rand_;
+  Envoy::Runtime::ScopedLoaderSingleton loader_;
+  Envoy::Filesystem::InstanceImplPosix file_system_;
+};
+
+TEST_F(ClientWorkerTest, BasicTest) {
+  ASSERT_EQ(std::this_thread::get_id(), thread_id_);
+
+  {
```
htuch: Maybe split these into 3 `TEST_F` cases.

nighthawk/source/client/client.cc
```diff
@@ -0,0 +1,224 @@
...
+    // TODO(oschaaf): output format, location, method, etc should be optionized.
+    std::cout << console_formatter.toString();
+    JsonOutputFormatterImpl json_formatter(time_source, *options_, merged_statistics,
+                                           merged_counters);
+    mkdir("measurements", 0777);
+    std::ofstream stream;
+    const int64_t epoch_seconds = time_source.systemTime().time_since_epoch().count();
+    std::string filename = fmt::format("measurements/{}.json", epoch_seconds);
+    stream.open(filename);
+    stream << json_formatter.toString();
```
htuch: This and `runWorkers` are pretty big functions, is there a reasonable way to factor out bits, e.g. output formatting here?

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  void SetUp() override {
+    // We fork the integration test fixture into a child process, to avoid conflicting
+    // runtimeloaders as both NH and the integration server want to own that and we can have only
+    // one.
+    pipe(fd_port);
+    pipe(fd_confirm);
+    pid_ = fork();
+    ASSERT(pid_ >= 0);
```
htuch: `RELEASE_ASSERT`

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,148 @@
...
+      // until that happens.
+      read(fd_confirm[0], &parent_message, sizeof(parent_message));
+      ASSERT(parent_message == 1234);
+      // The parent process closes fd_port when the test tears down. The read call blocks until it
+      // does that.
+      ASSERT(read(fd_port[0], &port_, sizeof(port_)) == -1);
+    } else if (pid_ > 0) {
+      ASSERT(read(fd_port[0], &port_, sizeof(port_)) > 0);
+      ASSERT(port_ > 0);
+      int foo = 1234;
```
htuch: 1234?

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,148 @@
...
+};
+
+std::string ClientTest::envoy_config;
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, ClientTest,
+                         testing::ValuesIn(Envoy::TestEnvironment::getIpVersionsForTest()),
+                         Envoy::TestUtility::ipTestParamsToString);
+
+TEST_P(ClientTest, NormalRun) {
+  if (pid_ > 0) {
```
htuch: Can't we kill off the `fork`ed gtest stuff at setup time, rather than in each test here?

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  void TearDown() override {
+    if (pid_ == 0) {
+      test_server_.reset();
+      fake_upstreams_.clear();
+      ares_library_cleanup();
+    }
+    close(fd_confirm[0]);
```
htuch: Kind of really not a fan of mixing gtest in parent and child, since we now have two interleaved programs. I really want something more like Python's subprocess. Envoy has a pretty good way of discovering its port even when binding admin and listener to 0. The admin port can be discovered via a CLI flag that dumps to a file, the listeners can be pulled from this. See https://github.com/envoyproxy/envoy/blob/d8e0f9c1cd3ad3ed6fc932d3033a6d29e62d3fa8/test/integration/hotrestart_test.sh#L71

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,148 @@
...
+      int port = lookupPort("listener_0");
+      int parent_message;
+      write(fd_port[1], &port, sizeof(port));
+      // The parent process writes to fd_confirm when it has read the port. This call to read blocks
+      // until that happens.
+      read(fd_confirm[0], &parent_message, sizeof(parent_message));
+      ASSERT(parent_message == 1234);
+      // The parent process closes fd_port when the test tears down. The read call blocks until it
+      // does that.
+      ASSERT(read(fd_port[0], &port_, sizeof(port_)) == -1);
```
htuch: `RELEASE_ASSERT` for all syscalls (including `close` ideally).

nighthawk/source/client/client.cc
```diff
@@ -0,0 +1,224 @@
...
+                            const std::vector<ClientWorkerPtr>& workers) const {
+  std::vector<StatisticPtr> merged_statistics;
+  StatisticPtrMap w0_statistics = workers[0]->statistics();
+  for (auto w0_statistic : w0_statistics) {
+    auto new_statistic = statistic_factory.create();
+    new_statistic->setId(w0_statistic.first);
+    merged_statistics.push_back(std::move(new_statistic));
+  }
+
+  for (auto& w : workers) {
```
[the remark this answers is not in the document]
oschaaf: No -- added comments to clarify

nighthawk/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+First, follow steps 1 and 2 over at [Quick start Bazel build for developers](https://github.com/envoyproxy/envoy/blob/master/bazel/README.md#quick-start-bazel-build-for-developers).
+
+## Building and testing Nighthawk
+```bash
+# build it. for best accuracy it is important to specify -c opt.
+bazel build -c opt //nighthawk:nighthawk_client
+
+# test it
+bazel test -c opt //nighthawk/test:nighthawk_test
```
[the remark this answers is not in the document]
oschaaf: yes, amended

nighthawk/test/client_test.cc
oschaaf: Made this a named constant.

nighthawk/test/client_test.cc
oschaaf: Yes, turns out we could. Fixed by using `SKIP_TEST` in the test setup.

nighthawk/test/client_test.cc
oschaaf: +1, it's on the tracking issue. I had not foreseen running into the singleton/runtime conflict issue with the integration test server.

nighthawk/source/client/client.cc
```diff
@@ -0,0 +1,224 @@
...
+  Envoy::ThreadLocal::InstanceImpl tls;
+  Envoy::Event::DispatcherPtr main_dispatcher(api.allocateDispatcher());
+  Uri uri = Uri::Parse(options_->uri());
+  tls.registerThread(*main_dispatcher, true);
+  try {
+    // TODO(oschaaf): verify with @htuch that ::Auto is the right default here.
+    // Also, this should be optionized.
+    uri.resolve(*main_dispatcher, Envoy::Network::DnsLookupFamily::Auto);
+  } catch (const UriException) {
+    tls.shutdownGlobalThreading();
```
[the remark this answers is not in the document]
oschaaf: Nice, yes we can, fixed.

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,147 @@
...
+        "nighthawk/test/test_data/benchmark_http_client_test_envoy.yaml"));
+    envoy_config = Envoy::TestEnvironment::substitute(envoy_config);
+  }
+
+  void SetUp() override {
+    // We fork the integration test fixture into a child process, to avoid conflicting
+    // runtimeloaders as both NH and the integration server want to own that and we can have only
+    // one. The plan is to move to python for this type of testing, so hopefully we can deprecate
+    // this test and it's peculiar setup with fork/pipe soon.
+    pipe(fd_port);
```
htuch: Can you wrap every single call, `pipe`, `write`, `read`, etc. with the expected return and a `RELEASE_ASSERT`? This would make me a lot more comfortable with the robustness here.

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,147 @@
...
+    pipe(fd_confirm);
+    pid_ = fork();
+    RELEASE_ASSERT(pid_ >= 0, "Fork failed");
+
+    const int kParentMessageId = 12341234;
```
htuch: Do we really need this?

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,147 @@
...
+    std::vector<const char*> argv;
+    for (const std::string& s : words) {
+      argv.push_back(s.c_str());
+    }
+    return std::make_unique<OptionsImpl>(argv.size(), argv.data());
+  }
+
+  int port_;
+  pid_t pid_;
+  int fd_port[2];
```
htuch: `fd_port_` (and same with other fields below)

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  int fd_confirm[2];
+  static std::string envoy_config;
```
htuch: Static non-PoD disallowed due to static initialization fiasco by the style guide, consider making this non-static, using const char[] or `CONSTRUCT_ON_FIRST_USE` instead.

nighthawk/test/client_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  argv.push_back(url.c_str());
+
+  Main program(argv.size(), argv.data());
+  EXPECT_TRUE(program.run());
+}
+
+TEST_P(ClientTest, BadRun) {
+  Main program(createOptionsImpl("foo --duration 1 --rps 1 https://unresolveable.host/"));
+  EXPECT_FALSE(program.run());
+}
```
htuch: This looks much nicer, thanks!

nighthawk/source/client/client.cc
```diff
@@ -0,0 +1,275 @@
...
+
+void Main::writeOutput(ProcessContext& context, const std::vector<StatisticPtr>& merged_statistics,
+                       const std::map<std::string, uint64_t>& merged_counters) const {
+  ConsoleOutputFormatterImpl console_formatter(context.time_system(), *options_, merged_statistics,
+                                               merged_counters);
+  // TODO(oschaaf): output format, location, method, etc should be optionized.
+  std::cout << console_formatter.toString();
+  JsonOutputFormatterImpl json_formatter(context.time_system(), *options_, merged_statistics,
+                                         merged_counters);
+  mkdir("measurements", 0777);
```
htuch: As with tests below, when doing syscalls, can you `RELEASE_ASSERT` on them (or check and handle errors, seems like this should be what we do in this one, but fine to leave as assert and TODO).

nighthawk/test/client_test.cc
oschaaf: removed, we now instead confirm that the process that runs the test actually read the correct port value that we send to it.

nighthawk/test/client_test.cc
```diff
@@ -29,49 +29,47 @@ namespace Client {
 class ClientTest : public Envoy::BaseIntegrationTest,
                    public testing::TestWithParam<Envoy::Network::Address::IpVersion> {
 public:
-  ClientTest() : Envoy::BaseIntegrationTest(GetParam(), realTime(), ClientTest::envoy_config) {}
+  ClientTest() : Envoy::BaseIntegrationTest(GetParam(), realTime(), readEnvoyConfiguration()) {}
 
-  static void SetUpTestCase() {
+  const std::string readEnvoyConfiguration() {
```
htuch: Nit: doesn't need to be `const` in return type.

nighthawk/test/client_test.cc
```diff
@@ -29,49 +29,47 @@ namespace Client {
...
       initialize();
       int port = lookupPort("listener_0");
       int parent_message;
-      write(fd_port[1], &port, sizeof(port));
-      // The parent process writes to fd_confirm when it has read the port. This call to read blocks
-      // until that happens.
-      read(fd_confirm[0], &parent_message, sizeof(parent_message));
-      RELEASE_ASSERT(parent_message == kParentMessageId, "Unexpected kParentMessageId value");
-      // The parent process closes fd_port when the test tears down. The read call blocks until it
+      write(fd_port_[1], &port, sizeof(port));
```
htuch: Nit: `RELEASE_ASSERT`

nighthawk/test/client_test.cc
```diff
@@ -104,12 +102,11 @@ class ClientTest : public Envoy::BaseIntegrationTest,
...
   pid_t pid_;
-  int fd_port[2];
-  int fd_confirm[2];
-  static std::string envoy_config;
+  int fd_port_[2];
+  int fd_confirm_[2];
 };
 
-std::string ClientTest::envoy_config;
+// std::string ClientTest::envoy_config;
```
htuch: ?

PR #68, opened 2019-04-04
Nighthawk output improvements

oschaaf:
- Use the proto as a base to compute CLI output in `ConsoleOutputFormatterImpl` to ensure generic behavior across output formats (it used to do that from the native objects).
- Add --format=[human|json|yaml] option
- Write per-worker statistics in json/yaml output
- Clearer CLI output
- Factorize OutputFormatter creation
- Stop using std::err. 
- Stop writing files, we probably want to do that elsewhere.

### DIFF
--- a/nighthawk/include/nighthawk/client/options.h
+++ b/nighthawk/include/nighthawk/client/options.h
@@ -27,6 +27,8 @@ public:
   virtual bool h2() const PURE;
   virtual std::string concurrency() const PURE;
   virtual std::string verbosity() const PURE;
+  virtual std::string output_format() const PURE;
+
   /**
    * Converts an Options instance to an equivalent CommandLineOptions instance in terms of option
    * values.
--- a/nighthawk/include/nighthawk/client/options.h
+++ b/nighthawk/include/nighthawk/client/options.h
@@ -19,7 +19,7 @@ class Options {
 public:
   virtual ~Options() {}
 
-  virtual uint64_t requests_per_second() const PURE;
+  virtual uint64_t requestsPerSecond() const PURE;
   virtual uint64_t connections() const PURE;
   virtual std::chrono::seconds duration() const PURE;
   virtual std::chrono::seconds timeout() const PURE;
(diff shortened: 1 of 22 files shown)


nighthawk/include/nighthawk/client/options.h
```diff
@@ -27,6 +27,8 @@ class Options {
   virtual bool h2() const PURE;
   virtual std::string concurrency() const PURE;
   virtual std::string verbosity() const PURE;
+  virtual std::string output_format() const PURE;
```
htuch: Nit: `outputFormat`

nighthawk/include/nighthawk/client/output_formatter.h
```diff
@@ -14,8 +14,13 @@ namespace Client {
 class OutputFormatter {
 public:
   virtual ~OutputFormatter() = default;
+  virtual void addResult(const std::string name, const std::vector<StatisticPtr>& statistics,
```
htuch: `absl::string_view name`? (or make it a ref?)

nighthawk/source/client/client.cc
```diff
@@ -187,18 +185,45 @@ class ProcessContext {
...
     for (auto& w : workers_) {
       w->start();
     }
-    bool ok = true;
     for (auto& w : workers_) {
       w->waitForCompletion();
       ok = ok && w->success();
     }
+
+    if (workers_.size() > 1) {
```
htuch: Nit: `if (!workers_.empty()) {`

htuch: Is this even needed, won't the `for` loop nop in the empty case?

oschaaf: This is comment-worthy, added.

nighthawk/include/nighthawk/client/options.h
oschaaf: renamed (+requests_per_second->requestsPerSecond for consistency)

nighthawk/include/nighthawk/client/output_formatter.h
oschaaf: done

nighthawk/source/client/client.cc
oschaaf: Comment added

oschaaf:
@htuch ready for another pass!

PR #69, opened 2019-04-05
 Nighthawk: add connection prefetching feature

oschaaf:
Enabling connection prefetching will cause the client to open the
requested amount of connections to the benchmark target before performing
the actual benchmark.

### DIFF
--- a/nighthawk/source/client/benchmark_client_impl.cc
+++ b/nighthawk/source/client/benchmark_client_impl.cc
@@ -33,11 +33,12 @@ BenchmarkClientHttpImpl::BenchmarkClientHttpImpl(Envoy::Api::Api& api,
                                                  Envoy::Stats::Store& store,
                                                  StatisticPtr&& connect_statistic,
                                                  StatisticPtr&& response_statistic, const Uri& uri,
-                                                 bool use_h2)
+                                                 bool use_h2, bool prefetch_connections)
     : api_(api), dispatcher_(dispatcher), store_(store),
       scope_(store_.createScope("client.benchmark.")),
       connect_statistic_(std::move(connect_statistic)),
-      response_statistic_(std::move(response_statistic)), use_h2_(use_h2), uri_(uri),
+      response_statistic_(std::move(response_statistic)), use_h2_(use_h2),
+      prefetch_connections_(prefetch_connections), uri_(uri),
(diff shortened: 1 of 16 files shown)


nighthawk/source/client/benchmark_client_impl.h
```diff
@@ -43,13 +43,20 @@ struct BenchmarkClientStats {
   ALL_BENCHMARK_CLIENT_STATS(GENERATE_COUNTER_STRUCT)
 };
 
+class PrefetchablePool {
```
htuch: Does this belong as a pure `include` interface?

nighthawk/source/client/benchmark_client_impl.cc
```diff
@@ -124,11 +151,15 @@ void BenchmarkClientHttpImpl::initialize(Envoy::Runtime::Loader& runtime) {
...
+                                     options);
   } else {
-    pool_ = std::make_unique<Envoy::Http::Http1::ProdConnPoolImpl>(
-        dispatcher_, host, Envoy::Upstream::ResourcePriority::Default, options);
+    pool_ = std::make_unique<H1Pool>(dispatcher_, host, Envoy::Upstream::ResourcePriority::Default,
+                                     options);
+  }
+
+  if (prefetch_connections_) {
+    prefetchPoolConnections();
```
htuch: IIUC, what this does is allow you to prefetch at the start of the benchmark. This seems useful. Is there any value on continuing to also prefetch when connections are dropped (due to exceeding max requests)?

oschaaf: I think that doing that would add value, maybe as a separate flag; we would make sure we mostly have the same number of connections open throughout the test. As the number of open connections is something that may influence server performance all by itself, it might be interesting.

nighthawk/source/client/benchmark_client_impl.h
oschaaf: Yes, let me add that.

nighthawk/source/client/benchmark_client_impl.cc
htuch: Ack, maybe put it on the backlog and we can resolve.

oschaaf: Added to the backlog

oschaaf:
This moved to https://github.com/envoyproxy/nighthawk/pull/9, closing this here.

PR #70, opened 2019-04-09
Basic test server

oschaaf:
Initial stab at a test origin for Nighthawk. Based on Envoy's libraries, like the client.

Current functionality is very simple: the request headers can be used to indicate how many bytes the response body should have. The response consist of repeating the character `a`.

Looking at the test coverage report, the code in this PR isn't perfectly covered yet, but it I feel its state is good enough to start reviewing it.

For reference, an example config that combines it with the `Fault-filter` to induce delayed responses: https://gist.github.com/oschaaf/1fa4b09ca2d70acbda60c2f2464b3a80

### DIFF
--- a/nighthawk/source/server/http_test_server_filter.cc
+++ b/nighthawk/source/server/http_test_server_filter.cc
@@ -36,10 +36,8 @@ HttpTestServerDecoderFilter::decodeHeaders(Envoy::Http::HeaderMap& headers, bool
       Envoy::MessageUtil::loadFromJson(request_config_header->value().c_str(), json_config);
       base_config.MergeFrom(json_config);
       Envoy::MessageUtil::validate(base_config);
-    } catch (Envoy::EnvoyException e_exception) {
-      error_message = e_exception.what();
-    } catch (Envoy::ProtoValidationException p_exception) {
-      error_message = p_exception.what();
+    } catch (Envoy::EnvoyException exception) {
+      error_message = exception.what();
     }
   }
 
--- a/nighthawk/source/server/http_test_server_filter.cc
+++ b/nighthawk/source/server/http_test_server_filter.cc
@@ -3,35 +3,35 @@
 
 #include "absl/strings/numbers.h"
 
-#include "http_test_origin_filter.h"
+#include "http_test_server_filter.h"
 
 #include "envoy/server/filter_config.h"
 
 namespace Nighthawk {
 namespace Server {
 
-HttpTestOriginDecoderFilterConfig::HttpTestOriginDecoderFilterConfig(
-    const nighthawk::server::TestOrigin& proto_config)
+HttpTestServerDecoderFilterConfig::HttpTestServerDecoderFilterConfig(
+    const nighthawk::server::TestServer& proto_config)
     : key_(proto_config.key()), val_(proto_config.val()) {}
 
-HttpTestOriginDecoderFilter::HttpTestOriginDecoderFilter(
-    HttpTestOriginDecoderFilterConfigSharedPtr config)
+HttpTestServerDecoderFilter::HttpTestServerDecoderFilter(
+    HttpTestServerDecoderFilterConfigSharedPtr config)
     : config_(config) {}
 
-HttpTestOriginDecoderFilter::~HttpTestOriginDecoderFilter() {}
+HttpTestServerDecoderFilter::~HttpTestServerDecoderFilter() {}
 
-void HttpTestOriginDecoderFilter::onDestroy() {}
+void HttpTestServerDecoderFilter::onDestroy() {}
 
-Envoy::Http::LowerCaseString HttpTestOriginDecoderFilter::headerKey() const {
+Envoy::Http::LowerCaseString HttpTestServerDecoderFilter::headerKey() const {
   return Envoy::Http::LowerCaseString(config_->key());
 }
 
-const std::string& HttpTestOriginDecoderFilter::headerValue() const { return config_->val(); }
+const std::string& HttpTestServerDecoderFilter::headerValue() const { return config_->val(); }
 
 Envoy::Http::FilterHeadersStatus
-HttpTestOriginDecoderFilter::decodeHeaders(Envoy::Http::HeaderMap& headers, bool) {
+HttpTestServerDecoderFilter::decodeHeaders(Envoy::Http::HeaderMap& headers, bool) {
   const auto response_size_header =
-      headers.get(TestOrigin::HeaderNames::get().TestOriginResponseSize);
+      headers.get(TestServer::HeaderNames::get().TestServerResponseSize);
   const int max = 1024 * 1024 * 4;
   int response_size = -1;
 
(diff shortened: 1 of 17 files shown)


nighthawk/source/server/http_test_server_filter.cc
```diff
@@ -0,0 +1,83 @@
...
+HttpTestServerDecoderFilter::decodeHeaders(Envoy::Http::HeaderMap& headers, bool) {
+  const auto request_config_header = headers.get(TestServer::HeaderNames::get().TestServerConfig);
+  nighthawk::server::TestServer base_config = config_->server_config();
+
+  bool error = false;
+  if (request_config_header != nullptr) {
+    try {
+      nighthawk::server::TestServer json_config;
+      Envoy::MessageUtil::loadFromJson(request_config_header->value().c_str(), json_config);
+      base_config.MergeFrom(json_config);
```
htuch: Will merge always be sufficient? Might there be some thing you want do (e.g. clear a response header) that can't be done easily via merge semantics?

nighthawk/source/server/http_test_server_filter.cc
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  // TODO(oschaaf): figure out how to get _cc_validate to be around in BUILD, and
+  // do something like this. Enable relevant tests when done.
+  // error = base_config.Validate();
+
+  if (error) {
+    decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(500),
+                                       "test-server didn't understand the request", nullptr,
```
htuch: It might be helpful to include the error message inside the `EnvoyException` above in the response body.

nighthawk/source/server/http_test_server_filter.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
+
+package nighthawk.server;
+
+import "google/protobuf/wrappers.proto";
+import "validate/validate.proto";
+
+// TODO(oschaaf): HeaderValue(Option) are copied from 
+// import "envoy/api/v2/core/base.proto";
+// Unfortunately, visibility aspects seem to prevent re-use as of writing this.
```
htuch: Let's chat on Slack, I think there are ways around this.

nighthawk/source/server/http_test_server_filter.proto
```diff
@@ -0,0 +1,40 @@
...
+message HeaderValueOption {
+  // Header name/value pair that this option applies to.
+  HeaderValue header = 1 [(validate.rules).message.required = true];
+
+  // Should the value be appended? If true (default), the value is appended to
+  // existing values.
+  google.protobuf.BoolValue append = 2;
+}
+
+message TestServer {
```
htuch: I would probably rename this as `ResponseOptions`. Should this proto go somewhere outside of source, e.g. into `nighthawk/proto` or `nighthawk/api/server`? You'll be building up a number of protos that define the NH APIs in the future.

nighthawk/source/server/http_test_server_filter.proto
```diff
@@ -0,0 +1,40 @@
...
+    repeated HeaderValueOption response_headers = 1;
+    oneof oneof_response_size {
```
htuch: What other things do you anticipate being `oneof` here?

nighthawk/source/server/http_test_server_filter_config.cc
```diff
@@ -0,0 +1,58 @@
...
+
+private:
+  Envoy::Http::FilterFactoryCb createFilter(const nighthawk::server::TestServer& proto_config,
+                                            Envoy::Server::Configuration::FactoryContext&) {
+    Nighthawk::Server::HttpTestServerDecoderFilterConfigSharedPtr config =
+        std::make_shared<Nighthawk::Server::HttpTestServerDecoderFilterConfig>(
+            Nighthawk::Server::HttpTestServerDecoderFilterConfig(proto_config));
+
+    return [config](Envoy::Http::FilterChainFactoryCallbacks& callbacks) -> void {
+      auto filter = new Nighthawk::Server::HttpTestServerDecoderFilter(config);
```
htuch: Nit: `auto*`

nighthawk/test/server/http_test_server_filter_integration_test.cc
```diff
@@ -0,0 +1,262 @@
...
+
+    dispatcher->run(Envoy::Event::Dispatcher::RunType::Block);
+    return response;
+  }
+
+  void testWithResponseSize(int response_size, bool expect_header = true) {
+    Envoy::BufferingStreamDecoderPtr response = makeSingleRequest(
+        lookupPort("http"), "GET", "/", "", downstream_protocol_, version_, "foo.com", "",
+        [response_size](Envoy::Http::HeaderMapImpl& request_headers) {
+          const std::string header_config = fmt::format("{{response_size:{}}}", response_size);
```
htuch: This is fine for such a simple message, if it gets more complicated I'd suggest building up the proto and then converting it to JSON.

nighthawk/source/server/http_test_server_filter.proto
oschaaf: Yes, that makes sense, done. Also added moving the other proto's out to the backlog.

nighthawk/source/server/http_test_server_filter.cc
htuch: Yes, this would work, good idea.

nighthawk/source/server/http_test_server_filter.proto
htuch: I think you can use a wrapped type, e.g. `google.protobuf.UInt32` here to avoid the use of `oneof`, it'll be a bit more intuitive.

nighthawk/source/server/http_test_server_filter.cc
```diff
@@ -0,0 +1,83 @@
...
+              direct_response_headers.remove(lower_case_key);
+            }
+            direct_response_headers.addCopy(lower_case_key, header.value());
+          }
+        },
+        absl::nullopt);
+  } else {
+    decoder_callbacks_->sendLocalReply(
+        static_cast<Envoy::Http::Code>(500),
+        fmt::format("test-server didn't understand the request: {}", error_message), nullptr,
```
htuch: Is this tested?

nighthawk/test/server/http_test_server_filter_integration_test.cc
```diff
@@ -0,0 +1,262 @@
...
+                                header_config);
+      });
+  ASSERT_TRUE(response->complete());
+  EXPECT_STREQ("200", response->headers().Status()->value().c_str());
+  EXPECT_STREQ("bar2",
+               response->headers().get(Envoy::Http::LowerCaseString("foo"))->value().c_str());
+  EXPECT_EQ(std::string(0, 'a'), response->body());
+}
+
+// TODO(oschaaf): maybe add a couple more tests for appending headers.
```
htuch: If the header merge/append logic is factored out into its own module, we can just do unit tests as well.

nighthawk/source/server/http_test_server_filter.cc
oschaaf: Yes, this code-path is hit in tests, checking we get a `500` response-code. Not checking the specific error messages though.

nighthawk/test/server/http_test_server_filter_integration_test.cc
oschaaf: Sure; would you like to see that in this PR? Or would you be ok with doing that in a follow up?

nighthawk/source/server/http_test_server_filter.proto
oschaaf: I tried to use that wrapped type earlier, but it didn't work out. Retracing my steps and trying again, `TestZeroLengthRequest` starts failing right away again.

nighthawk/source/server/http_test_server_filter.cc
htuch: Fair enough; it's probably also reasonable to verify the error as it's Envoy generated and should be relatively stable.

nighthawk/test/server/http_test_server_filter_integration_test.cc
htuch: I reckon let's do it here since it's probably not a ton of code, but up to you as well.

oschaaf:
Closing; this moved to https://github.com/envoyproxy/nighthawk/pull/8

PR #73, opened 2020-10-12
[salvo] Initial commit for benchmark abstraction framework

abaptiste:
This is the initial 'official' commit for the Salvo tool.  This aims to abstract the execution of nighthawk to benchmark a given envoy version. 

See [this issue](https://github.com/envoyproxy/envoy/issues/11266) for some background.  The two design docs for this project are referenced here.

In this commit, salvo is placed into a separate directory of the envoy-perf repository, and is referenced from the main README.md.

cc: @oschaaf, @htuch, @landesherr

Signed-off-by: Alvin Baptiste <alvinsb@gmail.com>

Commit Message: [salvo] Initial commit for benchmark abstraction framework
Testing: Unit tests included,  Address as many pylint3 issues as feasible
[#Issue] https://github.com/envoyproxy/envoy/issues/11266


mattklein123:
@jmarantz @oschaaf would you be willing to do a lightweight review of this?

oschaaf:
Sure; the first questions that pop up for me are:

- Do we want to establish/modify checks in CI for Python conventions? (And maybe proto3 & bazel files too?). We hypothetically could copy over some things from Nighthawk and/or Envoy to create a starting point. 
- There are build tasks in `.circleci/config.yml`, can we leverage that in this PR? E.g it seems unit tests are provided in here, we could execute those in CI?

salvo/README.md
```diff
@@ -0,0 +1,73 @@
+# salvo
```
oschaaf, line 1: Consistency nit: should "salvo" be capitalized? (below there are instances of that)

salvo/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using [NightHawk](https://github.com/envoyproxy/nighthawk).
+
+## Example Control Documents
+
+The control document defines the data needed to excute a benchmark.  At the moment the dockerized scavenging benchmark is the only one supported.  To run the benchmark, create a file with the following example contents:
```
oschaaf, line 7: Maybe a small glossary that describes what we mean when we say things like "scavenging" is helpful to those new to this?

salvo/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+JSON Example:
+
+```json
+{
```
oschaaf, line 12: Probably out of scope, but maybe nice to track as something that would be nice-to-have in the future: These things have a tendency to get outdated; I wonder if we can generate it from the current proto definition.  E.g. in NH we have [this script](https://github.com/envoyproxy/nighthawk/blob/master/tools/update_cli_readme_documentation.sh) which generates cli options into the documentation.

salvo/README.md
```diff
@@ -0,0 +1,73 @@
...
+  envoyImage: "envoyproxy/envoy-dev:f61b096f6a2dd3a9c74b9a9369a6ea398dbe1f0f"
+```
+
+In both examples, the envoy image being tested is a specific hash.  This hash can be replaced with "latest" to test the most recently created image against the previous image built from the prior Envoys master commit.
+
+
+## Building Salvo
+
+```bash
+bazel build //:salvo
```
oschaaf, line 53: Yeah maybe this might be good to also add to `ci/do_ci.sh` to make sure that what we document also executes without error.

salvo/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Running Salvo
+
+```bash
+bazel-bin/salvo --job ~/test_data/demo_jobcontrol.yaml
+```
+
+## Testing Salvo
+
+```bash
+ bazel test //test:*
```
oschaaf, line 65: Similarly,  maybe this might be good to add to `ci/do_ci.sh` to make sure these tests run without failures in CI.

salvo/salvo.py
```diff
@@ -0,0 +1,69 @@
...
+
+    benchmark.execute()
+
+    return 0
+
+
+if __name__ == '__main__':
+    sys.exit(main())
+
+#  vim: set ts=4 sw=4 tw=0 et :
```
oschaaf, line 69: Typically we don't have formatting metadata at the code level (or at least I'm not used to it), in NH we use [an external style file](https://github.com/envoyproxy/nighthawk/blob/master/tools/.style.yapf) which gets applies across all code (there's  scripts to check and fix too).

salvo/src/lib/api/control.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+package salvo;
+
+import "src/lib/api/image.proto";
+import "src/lib/api/source.proto";
+import "src/lib/api/env.proto";
+
+
+message JobControl {
```
oschaaf, line 10: Can we add a comment that describes the purpose of the `JobControl` message?

oschaaf (changes requested):
awesome; flushing out a first round of comments; the PR is pretty large so I may need some more takes to get through it all, but overall I have a feeling it looks great. Before going into more depth, I think that establishing automated code-level conventions would be good as that helps smooth the review process by automating away part of it.

salvo/src/lib/api/control.proto
abaptiste: Absolutely.  I'll add a comment

salvo/README.md
abaptiste: I have a `#TODO` to auto generate an example job control document.  Certainly on the radar

salvo/README.md
abaptiste: Right!  I'll update this comment.  Somehow I am fixated on the 'scavenging' benchmark, this is really the 'fully dockerized' test.

salvo/README.md
abaptiste: Fixed

salvo/README.md
abaptiste: Looks like the test failures are due to dependency issues (pytest, and python3.5).   I'll look at addressing this in a subsequent PR.  We were targeting python3.6 as interpreter, but it appears that python3.5 is the newest version available in the `envoyproxy/envoy-build` container.

htuch: Are you sure you can't access something like Python 3.8 by using `python3.8` directly? I've had to do this in a few places in Envoy CI.

abaptiste:
I'll be working on making this change set a bit smaller.

salvo/README.md
abaptiste: I can look into doing this. I started digging into this again today and had not ventured into Envoy to see how it's handled there.

mum4k:
Hi @abaptiste, my name is Jakub and I am one of the maintainers of [Nighthawk](https://github.com/envoyproxy/nighthawk). Recently I have been working on collecting [requirements](https://docs.google.com/document/d/1mAma-ksRN0OIBInoZKUdjdaIhK2nTQxFnCujq2fKi4E/edit) for a benchmarking framework of Envoy and I was very excited to hear that you are ahead and have something ready for review. My hope is that we can cooperate towards our common goals.

It will be my pleasure to start reviewing this PR and I am sure we can gain some traction going forward. Sorry about the delay you have encountered here so far, I was otherwise occupied and couldn't get here sooner. I will start reading the code here today and expect to ask some questions later today or tomorrow. I expect to start the review at a high level and will need you help to work towards the details.

Meanwhile it would help me if you could read the linked requirements and highlight any areas of concern or questions you might have. Looking forward to working with you on this.

oschaaf (approved):
Thanks @abaptiste for following up and cutting down PR size. This looks great to me, and I have no more comments; I will defer to @mum4k for another review.

mattklein123:
@mum4k thanks a ton for helping out with this review! My suggestion is to do a high level pass and ship and iterate but I will defer to all of you on how to proceed. It would be great if everyone can also read everyone else's docs and we can converge on a single plan. Feel free to also use issues, milestones, etc. in this repo if that would be helpeful!

salvo/tools/.style.yapf
```diff
@@ -0,0 +1,6 @@
+# The Google Python styles can be found here: https://github.com/google/styleguide/blob/gh-pages/pyguide.md
+# TODO: Look into enforcing single vs double quote.
```
mum4k, line 2: Glad to see we are planning to follow the linked style guide and bonus points for including automated checks using flake8. This will make it much easier for us to maintain the code going forward.

mum4k (changes requested):
Thank you for starting this effort. Still reading the code, but sending the very first round of high level comments. They mostly focus on structure of the packages. I am planning to review the api, the implementation and the tests in separate review rounds in this order.

ci/do_ci.sh
```diff
@@ -1,4 +1,43 @@
...
+# Build the salvo framework
+function build_salvo() {
+  echo "Building Salvo"
+  pushd salvo
+  bazel build //:salvo
+  popd
+}
+ 
+# Test the salvo framework
+# TODO(abaptiste) Tests currently fail in CI, but pass locally.
```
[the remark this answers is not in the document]
abaptiste: Yes, I was able to work past this.  This comment can be removed.

salvo/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    name = "api",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/lib/api:schema_proto",
+    ],
+)
+
+py_library(
+    name = "lib",
+    visibility = ["//visibility:public"],
```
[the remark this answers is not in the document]
abaptiste: I am still a relative newb when it comes to bazel.   That said, this likely doesn't need to be publicly visible.  Let me do some more reading on this and I will update it.

salvo/tools/shell_utils.sh
```diff
@@ -0,0 +1,25 @@
+source_venv() {
```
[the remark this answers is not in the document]
abaptiste: I borrowed these from a different repo but I will add comments here.

salvo/test/BUILD
```diff
@@ -0,0 +1,36 @@
+licenses(["notice"])  # Apache 2
```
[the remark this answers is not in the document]
abaptiste: These are unit tests and not integration tests.  I'd prefer to iterate on this and save this particular change for a subsequent PR.

salvo/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    ":api",
+    ":lib",
+  ],
+)
+
+py_library(
+    name = "api",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/lib/api:schema_proto",
```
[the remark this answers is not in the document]
abaptiste: There's a few reasons I am placing the "api" directory in src.  Most of it is due to being able to share the  protobuf generated code between a local service and a remote service.

This is another place where I prefer to iterate. I've restructured the tests to be alongside the modules they exercise as you suggested.

Lastly, I removed the globs :)

salvo/src/lib/api/control.proto
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  // Define where we find all required sources
+  repeated SourceRepository source = 6;
+
+  // Define the names of all required docker images
+  DockerImages images = 7;
+
+  // Define the environment variables needed for the test
+  EnvironmentVars environment = 8;
```
mum4k, line 31: And here to `environment_vars`.

salvo/src/lib/api/docker_volume.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
mum4k, line 1: (fyi) This file is a good case for us not having a single BUILD target for all these proto files. None of the other proto files seem to be importing this one, so having a separate compilation unit for it makes sense from the code perspective.

salvo/src/lib/api/docker_volume.proto
```diff
@@ -0,0 +1,21 @@
...
+
+package salvo;
+
+// This message defines the properties for a given mount. It is used
+// to generate the dictionary specifying volumes for the Python Docker
+// SDK
+message VolumeProperties {
+  // Defines a list of properties governing the mount in the container
+  string bind = 1;
```
mum4k, line 10: This is probably lack of knowledge on my end, but the names "bind" and "mode" don't help me much to understand the fields. Can we include examples of what they may contain? Once I see that I may be able to suggest better names.

salvo/src/lib/api/docker_volume.proto
```diff
@@ -0,0 +1,21 @@
...
+
+  // Define whether the mount point is read-write or read-only
+  string mode = 2;
+}
+
+// This message defines the volume structure consumed by the command
+// to run a docker image.
+message Volume {
+  // Specify a map of volumes and their mount points for use in a container
+  map<string, VolumeProperties> volumes = 1;
```
mum4k, line 20: Do we have a good reason to use a map? Or could we just go with a repeated field of a message that includes both the string (I am guessing it is a name) and the volume properties?

Proto maps are a cool feature that helps to optimize code. I.e. we would use it if there are hundreds of volumes and we need to quickly look one up by name. If we are dealing with a much smaller set, we should not try to optimize prematurely since it comes with a cost. The serialization order of proto maps is undefined so this will make our life harder in any sort of unit tests and integration tests we will end up writing.

mum4k, line 20: The field name suggests that the volume contains a set of volumes? Is that the architecture?

Or is it that the volume contains a repeated set of volume properties? In which case we may want to call the field `volume_properties`.

salvo/src/lib/api/env.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package salvo;
+
+// Capture all Environment variables required for the benchmark
+message EnvironmentVars {
+  // Specify the IP version for tests
+  oneof envoy_ip_test_versions {
+    bool v4only = 1;
```
mum4k, line 9: For code readability consider longer names, since only the field name will appear in the code:

Example:
`ipv4_only`, `ipv6_only`, `dual_stack`

Or similar. Currently seeing the name `all` in code won't give much context.

salvo/src/lib/api/env.proto
```diff
@@ -0,0 +1,25 @@
...
+  string envoy_path = 4;
+
+  // Specify the output directory for nighthawk artifacts
+  string output_dir = 5;
+
+  // Specify the directory where external tests are located
+  string test_dir = 6;
+
+  // Additional environment variables that may be needed for operation
+  map<string, string> variables = 7;
```
mum4k, line 24: Same comment about proto maps here, unless we are optimizing, let's prefer repeated fields.

salvo/src/lib/api/env.proto
```diff
@@ -0,0 +1,25 @@
...
+// Capture all Environment variables required for the benchmark
+message EnvironmentVars {
+  // Specify the IP version for tests
+  oneof envoy_ip_test_versions {
+    bool v4only = 1;
+    bool v6only = 2;
+    bool all = 3;
+  }
+
+  // Controls whether envoy is placed between the nighthawk client and server
```
mum4k, line 14: Do we plan to have an execution mode where Envoy won't be started? Can you help me understand our plan around that?

(optional) Secondly it feels we are abusing field by giving its empty value a special meaning. It may be more explicit and readable if we have a separate explicit field, say a boolean called `include_envoy`.

salvo/src/lib/api/image.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package salvo;
+
+// Capture all docker images required for the benchmark. This object is examined
```
mum4k, line 5: Possible typo: Captures ...

salvo/src/lib/api/image.proto
```diff
@@ -0,0 +1,25 @@
...
+// first before evaluating the state of specified source locations
+message DockerImages {
+  // Determines whether required docker are used if they already exist. The
```
mum4k, line 8: This sentence doesn't read well, can we try to improve it?

salvo/src/lib/api/image.proto
```diff
@@ -0,0 +1,25 @@
...
+  // benchmark image and binary image must be specified. If this is set to false
+  // there must be a specified source location from which we build the image
```
mum4k, line 10: What do we mean by "specified source location"? How does one specify it?

salvo/src/lib/api/image.proto
```diff
@@ -0,0 +1,25 @@
...
+
+  // This should be implicit. Images are preferred over building from source. We
+  // build only if the image pull is not successful and sources are present
+  bool reuse_nh_images = 1;
```
mum4k, line 14: We seem to spell out `nighthawk` in most names except here. Can we change this name to `reuse_nighthawk_images` instead? It is a bit longer, but longer more explicit names help onboarding and improve code readability.

salvo/src/lib/api/image.proto
```diff
@@ -0,0 +1,25 @@
...
+package salvo;
+
+// Capture all docker images required for the benchmark. This object is examined
+// first before evaluating the state of specified source locations
+message DockerImages {
+  // Determines whether required docker are used if they already exist. The
+  // benchmark image and binary image must be specified. If this is set to false
+  // there must be a specified source location from which we build the image
+
+  // This should be implicit. Images are preferred over building from source. We
```
mum4k, line 12: I don't really understand the comment as it stands. "This should be implicit." sounds like a wish and I am not sure in what context we are making the wish and why it isn't under our control.

Maybe we can improve the comment by instead focusing on describing the meaning of the proto field?

salvo/src/lib/api/source.proto
```diff
@@ -0,0 +1,35 @@
...
+message SourceRepository {
+  // Specify whether this source location is Envoy or NightHawk
+  oneof identity {
+    bool envoy = 1;
+    bool nighthawk = 2;
+  }
+
+  // Specify the location of the source repository on disk. If specified
+  // this location is used to determine the origin url, branch, and commit
+  // hash. If not specified, the remaining fields must be populated
```
mum4k, line 15: Seems this would be a good application for a oneof. That way we can make the API self-explanatory when it comes to understanding which fields should be populated and which are mutually exclusive.

We can add two separate messages, e.g. LocalSourceLocation and RemoteSourceLocation. The LocalSourceLocation would only contain a single string field `directory_path`. The RemoteSourceLocation can contain all the otherfields. The parent message SourceRepository can then have a oneof between the two messages. This will also simplify a lot of the comments here, since we no longer need to say things like "This is ignored if...".

salvo/src/lib/api/source.proto
```diff
@@ -0,0 +1,35 @@
...
+  string location = 3;
+
+  // Specify the remote location of the repository. This is ignored if
+  // the source location is specified.
+  string url = 4;
```
mum4k, line 20: In the comments for these fields - can we include one valid example for each of the field values?

salvo/src/lib/api/source.proto
```diff
@@ -0,0 +1,35 @@
...
+
+  // Specify the local working branch.This is ignored if the source
+  // location is specified.
+  string branch = 5;
+
+  // Specify a commit hash if applicable. If not identified we will
+  // determine this from the source tree. We will also use this field
+  // to identify the corresponding NightHawk or Envoy image used for
+  // the benchmark
+  string hash = 6;
```
mum4k, line 30: Can we call this `commit_hash`?

salvo/src/lib/api/source.proto
```diff
@@ -0,0 +1,35 @@
...
+
+  // Internal use only. This field is used to specify the location of
```
mum4k, line 32: Can you help me understand the use case for this internal-only field? I wonder if we can get by without including it in the API.

salvo/tools/format_python_tools.sh
```diff
@@ -0,0 +1,39 @@
...
+# Because of conflict with the automatic fix format script, we ignore: 
+# E111 Indentation is not a multiple of four
+# E114 Indentation is not a multiple of four (comment)
+# E501 Line too long (82 > 79 characters)
+# E124 Closing bracket does not match visual indentation
+# E125 Continuation line with same indent as next logical line
+# E126 Continuation line over-indented for hanging indent
+
+# We ignore false positives because of what look like pytest peculiarities 
+# F401 Module imported but unused
```
mum4k, line 33: We didn't get to reviewing Python code yet, but these two feel like they may be result of how we are importing code / setting up the build system. It may be valuable for us to try and address them before we silence them which can result in accumulation of technical debt. Can you help me out by pointing to the locations where these warnings come up? I might be able to suggest a simple fix.

mum4k (changes requested):
Thank you for iterating on this, the BUILD system setup looks much easier to follow. I have a few more improvement suggestions on that front. I have also reviewed the documentation, most of the helper scripts and the proto API in this round.

I will start reviewing the Python code in the next round. Thank you for your patience, this is mostly taking this many rounds because the scope of this PR is fairly large. We can try to address that in future PRs by making them smaller. Happy to discuss the best approach, one option is to send separate PRs per each api / module / package. Reviewing and addressing comments on large PRs can sometimes feel like running a  marathon ;)

ci/do_ci.sh
```diff
@@ -1,4 +1,46 @@
...
+  popd
+}
+ 
+# Test the salvo framework
+function test_salvo() {
+  echo "Running Salvo unit tests"
+  pushd salvo
+  ./install_deps.sh
+
+  # TODO(abaptiste): Discover tests vs listing them individually
```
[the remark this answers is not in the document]
abaptiste: If listing the tests individually is fine, then I can leave this as is.  I would prefer to use the ellipses as you mentioned, but I would have to iterate on that to get bazel to cooperate.

salvo/README.md
```diff
@@ -0,0 +1,76 @@
...
```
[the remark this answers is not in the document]
abaptiste: Updated.  I will highlight the `do_ci.sh` script until we have a different method of executing all of the tests

salvo/src/lib/api/source.proto
abaptiste: In the case where we run the benchmark remotely, we generate a diff of any local changes and ship that to the remote service.  In the remote model, we can pull the source tree from github, checkout the current commit hash, and apply the patch.  This should replicate the user's changes remotely.  From there we would build an Envoy container for benchmarking.

We could probably remove this field, but we have not gotten to the point yet where we are using it.

salvo/tools/format_python_tools.sh
abaptiste: I copied these scripts from [nighthawk](https://github.com/envoyproxy/nighthawk/tree/master/tools) and they are largely unmodified.

Can we circle back to this because there are a few games we have to play with adding directories to the path because of the source layout and 'bazel'?    We can certainly tighten up the style and formatting checks.

salvo/src/lib/api/source.proto
abaptiste: Certainly.  I've made that change within the code itself. It makes sense here too.

salvo/src/lib/api/source.proto
abaptiste: Updated

salvo/src/lib/api/source.proto
abaptiste: Updated

salvo/src/lib/BUILD
```diff
@@ -0,0 +1,58 @@
+load("@rules_python//python:defs.bzl", "py_library", "py_test")
+
+licenses(["notice"])  # Apache 2
+
+py_library(
+    name = "api",
```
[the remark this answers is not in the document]
abaptiste: Ok, I'll move this.  We'll figure out how to make it work with other things later.

salvo/BUILD
```diff
@@ -0,0 +1,29 @@
...
+    srcs = ["salvo.py"],
+    srcs_version = "PY3",
+    deps = [
+        ":api",
+        ":lib",
+    ],
+)
+
+py_library(
+    name = "api",
```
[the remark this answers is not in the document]
abaptiste: I think this may be addressed with some other changes.  If not, feel free to un-resolve this thread.

ci/do_ci.sh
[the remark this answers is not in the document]
abaptiste: I'll include this update to the ci script in the next batch of changes.

mum4k (changes requested):
Thank you for cleaning up the build system setup. Looks like this PR still has a number of comments that weren't addressed yet (they are marked as "Outdated" due to the file moves, but their content still applies). Can you please look at them in the conmversation history and address them before we move forward?

Most of the comments that are still in flight are related to the API. We have already finished reviewing the documentation and helper scripts (format tools, download dependencies, repo setup). I have a proposal that may help us decrease the scope of this PR while moving forward. How would you feel about removing all the Python code and tests from this PR and merging what we have reviewed so far (after addressing the existing comments)? We could then immediately follow up with a set of smaller PRs, one per each Python library and its tests. The benefit would be that each of the smaller PRs would take less time in review and would make it easier for us to keep track of the conversation and changes.

Please let me know what you think.

salvo/src/lib/BUILD
```diff
@@ -0,0 +1,58 @@
...
+    visibility = ["//visibility:public"],
+)
+
+py_test(
+    name = "test_docker_image",
+    srcs = ["test_docker_image.py"],
+    srcs_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = [
+        "docker_lib",
```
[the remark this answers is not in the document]
abaptiste: Ack

ci/do_ci.sh
```diff
@@ -1,4 +1,44 @@
...
+
+# Avoid using "-x" here so that we do not leak credentials
+# in build logs
+set -e
+
+# Build the salvo framework
+function build_salvo() {
+  echo "Building Salvo"
+  pushd salvo
+  bazel build //:salvo
```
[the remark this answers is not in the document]
abaptiste: Ack

salvo/README.md
```diff
@@ -0,0 +1,78 @@
+# Salvo
+
+This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using [NightHawk](https://github.com/envoyproxy/nighthawk).
+
+## Example Control Documents
+
+The control document defines the data needed to excute a benchmark. At the moment, the fully dockerized benchmark is the only one supported. This benchmark discoveres user supplied tests for execution and uses docker images to run the tests. In the exampple below, the user supplied tests files are located in `/home/ubuntu/nighthawk_tests` and are mapped to a volume in the docker container.
```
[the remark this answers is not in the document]
abaptiste: fixed

salvo/README.md
```diff
@@ -0,0 +1,78 @@
...
+To run the benchmark, create a file with the following example contents:
+
+JSON Example:
+
+```json
+{
+  "remote": false,
+  "dockerizedBenchmark": true,
+  "images": {
+    "reuseNhImages": true,
```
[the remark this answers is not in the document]
abaptiste: Right.  Certainly doable.

salvo/src/lib/BUILD
```diff
@@ -0,0 +1,58 @@
...
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/lib/api:schema_proto",
+    ],
+)
+
+py_library(
+    name = "docker_lib",
+    data = [
+        "docker_image.py",
```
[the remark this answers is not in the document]
abaptiste: Fixed

salvo/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+
+licenses(["notice"])  # Apache 2
+
+py_binary(
+    name = "salvo",
+    srcs = ["src/salvo.py"],
```
mum4k, line 7: As discussed offline, the move of `salvo.py` into the `src/` directory was likely caused by a mistake I made in one of my comments. If we can, let's move it back to the `salvo/` directory.

salvo/src/lib/docker_volume.py
```diff
@@ -0,0 +1,57 @@
...
+from google.protobuf.json_format import (Error, MessageToJson)
+
+from api.docker_volume_pb2 import Volume, VolumeProperties
+
+log = logging.getLogger(__name__)
+
+
+class DockerVolume():
+
+  @staticmethod
```
mum4k, line 21: Instead of marking this method static, can we define it as a module level function (i.e. remove the class DockerVolume).

We should prefer standalone functions unless having a class is really necessary. We can test for that by identifying if the class either has interesting state that changes during its lifetime or if the code needs to maintain multiple instances of the class. If we don't have good reasons for either of these two, having a standalone method will result in simpler code.

The style guide also recommends avoiding staticmethod:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#217-function-and-method-decorators

salvo/src/salvo.py
```diff
@@ -0,0 +1,63 @@
...
+
+import argparse
+import logging
+import os
+import site
+import sys
+
+# Run in the actual bazel directory so that the sys.path
+# is setup correctly
+if os.path.islink(sys.argv[0]):
```
mum4k, line 11: Can you help me understand this path manipulation and the `site.addsitedir("src")` call below?

I wouldn't expect this to be necessary in a bazel environment. We should be able to delete all these manipulations from here and the other files and just go with an import rooted at the beginning of the WORKSPACE:

`from src.lib.job_control_loader import load_control_doc`

I have tested this locally by removing all the manipulation and executing the `salvo` target which seems to have worked at least to a point that it imported all libraries correctly. Do we need to keep this or can we do without it?

mum4k (changes requested):
Thank you for the offline discussion @abaptiste. As we agreed, we will keep the Python code in this PR and achieve greater velocity by instead switching the focus of this review to architecture. This will help us move faster while giving us the opportunity to iterate on the code in future changes.

With that said, this review round represents the last set of comments I have at that level. Thank you for your patience with this process and looking forward to the next PRs. I will tag one of the maintainers here for merging after these comments are addressed.

salvo/BUILD
abaptiste: I had a merge conflict and I didn't catch this.  I'll fix and update

ci/do_ci.sh
```diff
@@ -1,4 +1,44 @@
...
+  popd
+}
+ 
+# Test the salvo framework
+function test_salvo() {
+  echo "Running Salvo unit tests"
+  pushd salvo
+  ./install_deps.sh
+
+  bazel test //src/lib/...
```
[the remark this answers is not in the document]
abaptiste: Updated

salvo/src/lib/docker_volume.py
abaptiste: Updated.

salvo/src/salvo.py
abaptiste: This is an artifact of us getting salvo to run in a cloud service.  We noticed that bazel would symlink files to the original location so, the current working directory would be the local source location and not `bazel-bin` where we expect.  This was true for the python generated code for the api.  We needed these files packaged in a library and were not able to accomplish that without the working directory manipulation.

I can remove it here, but there are other instance where we need it to remain.

abaptiste:
@mum4k I pushed the last set of changes.  Thanks for your reviews!

mum4k (dismissed):
Thank you for addressing the comments @abaptiste.

@htuch this PR is good to go in, please take a look.

salvo/api/docker_volume.proto
```diff
@@ -0,0 +1,22 @@
...
+// This message defines the properties for a given mount. It is used
+// to generate the dictionary specifying volumes for the Python Docker
+// SDK.  This message is not constructed by the user
+message VolumeProperties {
+  // Specified the destination mount path in the benchmark container
+  string bind = 1;
+
+  // Define whether the mount point is read-write or read-only
+  // eg: 'rw' or 'ro'
+  string mode = 2;
```
htuch, line 14: Probably should be an enum, but don't feel super strong.

salvo/api/env.proto
```diff
@@ -0,0 +1,27 @@
...
+
+package salvo;
+
+// Capture all Environment variables required for the benchmark
+message EnvironmentVars {
+  // Specify the IP version for tests
+  oneof envoy_ip_test_versions {
+    bool v4only = 1;
+    bool v6only = 2;
+    bool all = 3;
```
htuch, line 11: Probably should an enum.

salvo/api/source.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
+
+package salvo;
+
+// Capture the location of sources needed for the benchmark
+message SourceRepository {
+  // Specify whether this source location is Envoy or NightHawk
+  oneof identity {
+    bool envoy = 1;
+    bool nighthawk = 2;
```
htuch, line 10: Enum please.

salvo/src/lib/docker_volume.py
```diff
@@ -0,0 +1,53 @@
...
+  # Setup the output directory
+  properties = VolumeProperties()
+  properties.bind = output_dir
+  properties.mode = 'rw'
+  volume_cfg.volumes[output_dir].CopyFrom(properties)
+
+  # Setup the test directory
+  if test_dir:
+    properties = VolumeProperties()
+    properties.bind = '/usr/local/bin/benchmarks/benchmarks.runfiles/nighthawk/benchmarks/external_tests/'
```
htuch, line 42: What is this hardcoded path?

salvo/src/lib/job_control_loader.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def load_control_doc(filename):
+  """
+    Return a JobControl object from the identified filename
+    """
+  contents = None
+
+  # Try loading the contents based on the file extension
+  if re.match(r'.*\.json', filename):
```
htuch, line 57: `filename.endswith('.json')` etc. below

salvo/src/lib/test_docker_image.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_run_image():
+  """Test executing a command in an image"""
+
+  if not os.path.exists("/var/run/docker.sock"):
+    pytest.skip("Skipping docker test since no socket is available")
+
+  env = ['key1=val1', 'key2=val2']
+  cmd = ['uname', '-r']
+  image_name = 'oschaaf/benchmark-dev:latest'
```
htuch, line 35: How come Otto's image is hardcoded?

salvo/src/lib/test_job_control_loader.py
```diff
@@ -0,0 +1,262 @@
...
+      assert source.source_url == "https://github.com/envoyproxy/nighthawk.git"
+      assert source.branch == "master"
+      assert not source.commit_hash
+      saw_nighthawk = True
+
+    elif source.envoy:
+      assert source.source_path == "/home/ubuntu/envoy"
+      assert not source.source_url
+      assert source.branch == "master"
+      assert source.commit_hash == "e744a103756e9242342662442ddb308382e26c8b"
```
htuch, line 78: What's the plan to deal with these hardcoded hashes? Should they be in some configuration file?

htuch (commented):
Since @mum4k is good with this, I think it would be great to ship and land. I had a quick scan and noticed a few nits that would be good to sort out, happy to merge when they are fixed.

salvo/src/lib/test_job_control_loader.py
abaptiste: These are just string literals with no meaning.  This could be "foo" and work the same.

salvo/src/lib/test_docker_image.py
abaptiste: It was available in dockerhub and I used it to validate that the docker module could actually pull an image and execute a command.  The actual image here is moot and could be any valid docker image.

salvo/src/lib/job_control_loader.py
abaptiste: Certainly.  Will update.

abaptiste:
@htuch This is ready for you again.  Thanks!

htuch:
@abaptiste looks good, but what about the enums?

salvo/api/env.proto
abaptiste: Updated

salvo/api/source.proto
abaptiste: Updated

salvo/src/lib/docker_volume.py
abaptiste: This is the location in the container that the external tests are mapped to.  The path is used from [this](https://github.com/envoyproxy/nighthawk/tree/master/benchmarks#example-fully-dockerized-flow) example.

salvo/api/docker_volume.proto
abaptiste: Enum it is.

abaptiste:
@htuch Sorry about the enums.  No excuses.  Updated the PR.  Please review at your leisure.

htuch (approved):
Thanks, excited to see this land!

oschaaf:
Exciting indeed, awesome!

abaptiste:
Thanks folks!

PR #74, opened 2020-11-12
Fully dockerized flow skeleton

abaptiste:
This change adds a skeleton for the fully dockerized benchmark flow.

This benchmark method uses docker images for  Envoy, the NightHawk benchmark framework, and the NightHawk binaries.  

Users specify a location where external tests are mapped into the benchmark container and executed.

cc: @mum4k, @oschaaf, @htuch 

Signed-off-by: Alvin Baptiste <alvinsb@gmail.com>

### DIFF
--- a/salvo/api/env.proto
+++ b/salvo/api/env.proto
@@ -6,10 +6,10 @@ package salvo;
 message EnvironmentVars {
   // Specify the IP version for tests
   enum TestIpVersion {
-    UNSPECIFIED = 0;
-    V4ONLY = 1;
-    V6ONLY = 2;
-    ALL = 3;
+    IPV_UNSPECIFIED = 0;
+    IPV_V4ONLY = 1;
+    IPV_V6ONLY = 2;
+    IPV_ALL = 3;
   }
 
   TestIpVersion test_version = 1;
--- a/salvo/api/env.proto
+++ b/salvo/api/env.proto
@@ -6,9 +6,10 @@ package salvo;
 message EnvironmentVars {
   // Specify the IP version for tests
   enum TestIpVersion {
-    V4ONLY = 0;
-    V6ONLY = 1;
-    ALL = 2;
+    UNSPECIFIED = 0;
+    V4ONLY = 1;
+    V6ONLY = 2;
+    ALL = 3;
   }
 
   TestIpVersion test_version = 1;
--- a/salvo/api/source.proto
+++ b/salvo/api/source.proto
@@ -7,8 +7,9 @@ message SourceRepository {
 
   // Specify whether this source location is Envoy or NightHawk
   enum SourceIdentity {
-    ENVOY = 0;
-    NIGHTHAWK = 1;
+    UNSPECIFIED = 0;
+    ENVOY = 1;
+    NIGHTHAWK = 2;
   }
 
   SourceIdentity identity = 1;
--- a/salvo/api/source.proto
+++ b/salvo/api/source.proto
@@ -7,9 +7,9 @@ message SourceRepository {
 
   // Specify whether this source location is Envoy or NightHawk
   enum SourceIdentity {
-    UNSPECIFIED = 0;
-    ENVOY = 1;
-    NIGHTHAWK = 2;
+    SRCID_UNSPECIFIED = 0;
+    SRCID_ENVOY = 1;
+    SRCID_NIGHTHAWK = 2;
   }
 
   SourceIdentity identity = 1;
--- a/salvo/src/lib/BUILD
+++ b/salvo/src/lib/BUILD
@@ -2,12 +2,15 @@ load("@rules_python//python:defs.bzl", "py_library", "py_test")
 
 licenses(["notice"])  # Apache 2
 
+package(
+    default_visibility = ["//:__subpackages__"],
+)
+
 py_library(
     name = "docker_image",
     data = [
         "docker_image.py",
     ],
-    visibility = ["//visibility:public"],
 )
 
 py_library(
@@ -15,7 +18,6 @@ py_library(
     data = [
         "docker_volume.py",
     ],
-    visibility = ["//visibility:public"],
 )
 
 py_library(
@@ -23,7 +25,13 @@ py_library(
     data = [
         "job_control_loader.py",
     ],
-    visibility = ["//visibility:public"],
+)
+
+py_library(
+    name = "constants",
+    data = [
+        "constants.py",
+    ],
 )
 
 py_library(
@@ -49,6 +57,7 @@ py_test(
     srcs = ["test_job_control_loader.py"],
     srcs_version = "PY3",
     deps = [
+        ":constants",
         ":job_control",
         "//api:schema_proto",
     ],
--- a/salvo/src/lib/BUILD
+++ b/salvo/src/lib/BUILD
@@ -18,6 +18,9 @@ py_library(
     data = [
         "docker_volume.py",
     ],
+    deps = [
+        ":constants",
+    ],
 )
 
 py_library(
@@ -46,6 +49,7 @@ py_test(
     srcs = ["test_docker_image.py"],
     srcs_version = "PY3",
     deps = [
+        ":constants",
         ":docker_image",
         ":docker_volume",
         "//api:schema_proto",
--- a/salvo/src/lib/BUILD
+++ b/salvo/src/lib/BUILD
@@ -7,6 +7,7 @@ py_library(
     data = [
         "docker_image.py",
     ],
+    visibility = ["//visibility:public"],
 )
 
 py_library(
@@ -14,6 +15,7 @@ py_library(
     data = [
         "docker_volume.py",
     ],
+    visibility = ["//visibility:public"],
 )
 
 py_library(
--- a/salvo/src/lib/benchmark/BUILD
+++ b/salvo/src/lib/benchmark/BUILD
@@ -1,0 +1,28 @@
+load("@rules_python//python:defs.bzl", "py_library", "py_test")
+
+licenses(["notice"])  # Apache 2
+
+py_library(
+    name = "benchmark",
+    data = [
+        "base_benchmark.py",
+        "fully_dockerized_benchmark.py"
+    ],
+    deps = [
+        "//src/lib:docker_image",
+        "//src/lib:docker_volume",
+    ],
+    visibility = [ "//visibility:public" ]
+)
+
+py_test(
+  name = "test_fully_dockerized_benchmark",
+  srcs = [ "test_fully_dockerized_benchmark.py" ],
+  srcs_version = "PY3",
+  deps = [
+      "//api:schema_proto",
+      "//src/lib:docker_image",
+      ":benchmark"
+  ],
+)
+
--- a/salvo/src/lib/benchmark/BUILD
+++ b/salvo/src/lib/benchmark/BUILD
@@ -2,6 +2,10 @@ load("@rules_python//python:defs.bzl", "py_library", "py_test")
 
 licenses(["notice"])  # Apache 2
 
+package(
+    default_visibility = ["//:__subpackages__"],
+)
+
 py_library(
     name = "benchmark",
     data = [
@@ -12,7 +16,6 @@ py_library(
         "//src/lib:docker_image",
         "//src/lib:docker_volume",
     ],
-    visibility = [ "//visibility:public" ]
 )
 
 py_test(
--- a/salvo/src/lib/benchmark/base_benchmark.py
+++ b/salvo/src/lib/benchmark/base_benchmark.py
@@ -22,6 +22,7 @@ def get_docker_volumes(output_dir, test_dir=''):
   """
   return docker_volume.generate_volume_config(output_dir, test_dir)
 
+
 class BaseBenchmark(object):
   """Base Benchmark class with common functions for all invocations."""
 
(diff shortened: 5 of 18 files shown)


salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -0,0 +1,68 @@
...
+import src.lib.docker_image as docker_image
+import src.lib.docker_volume as docker_volume
+
+log = logging.getLogger(__name__)
+"""
+Base Benchmark class with common functions for all invocations
+"""
+
+
+def get_docker_volumes(output_dir, test_dir=None):
```
mum4k, line 17: A question on privacy - are all of these functions and method meant to be public? (will they be used from outside of this package?

If some of them aren't we should prefer to make them private by prepending an underscore to their name. Private methods make it easier to change code around in the future.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#3162-naming-conventions

mum4k, line 17: Assuming the test_dir is meant to be a string, can we try to also keep its default value within the same object type? Even though Python isn't type safe, it is good to follow the generic type safety rules to avoid unexpected programming errors, like when someone tries to call a string method on a NoneType.

```py
def get_docker_volumes(output_dir, test_dir=''):
```

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -0,0 +1,68 @@
+"""
```
mum4k, line 1: (optional, food for thought) How would you feel about using pytype annotations in code in this repository? They tend to help a lot by being able to statically discover most common programming errors in Python.

Let me know if this is something you would be willing to explore, I can provide some hints.

Some background:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#221-type-annotated-code

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -0,0 +1,37 @@
...
+import logging
+
+from src.lib.benchmark.base_benchmark import BaseBenchmark
+
+log = logging.getLogger(__name__)
+
+
+class Benchmark(BaseBenchmark):
+
+  def __init__(self, **kwargs):
```
mum4k, line 17: What arguments are we expecting to be passing here and can we name them instead of using `kwargs`? Listing the arguments specifically would help us to think about code structure and would make the code more readable.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -0,0 +1,37 @@
...
+    super(Benchmark, self).__init__(**kwargs)
+
+  def validate(self):
```
mum4k, line 20: Since this is a public method, are we expecting the callers to call it specifically? Is that the right approach or should it be a private method that we always call from say `execute_benchmark`?

Other way of asking the same is - are there any situations where the caller would want to execute without validating? If we call it internally we may get better safety (impossible to execute without validation) and smaller API surface.

mum4k (changes requested):
Thank you for splitting the API into its own PR and keeping the scope limited.

Added a few comments and some questions. As discussed previously - please tag me on the PR specifically when all comments will be responded to and you want me to take another look.

salvo/api/env.proto
```diff
@@ -6,9 +6,10 @@ package salvo;
 message EnvironmentVars {
   // Specify the IP version for tests
   enum TestIpVersion {
-    V4ONLY = 0;
-    V6ONLY = 1;
-    ALL = 2;
+    UNSPECIFIED = 0;
+    V4ONLY = 1;
+    V6ONLY = 2;
+    ALL = 3;
```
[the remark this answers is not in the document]
abaptiste: This makes sense.  I will update

salvo/api/source.proto
```diff
@@ -7,8 +7,9 @@ message SourceRepository {
 
   // Specify whether this source location is Envoy or NightHawk
   enum SourceIdentity {
-    ENVOY = 0;
-    NIGHTHAWK = 1;
+    UNSPECIFIED = 0;
```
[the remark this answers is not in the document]
abaptiste: ack

salvo/src/lib/BUILD
```diff
@@ -7,13 +7,15 @@ py_library(
     data = [
         "docker_image.py",
     ],
+    visibility = ["//visibility:public"],
```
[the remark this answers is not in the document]
abaptiste: Let me try this.  I tried some variations based on the docs and got errors about circular dependencies.

salvo/src/lib/benchmark/BUILD
```diff
@@ -0,0 +1,28 @@
...
+    name = "benchmark",
+    data = [
+        "base_benchmark.py",
+        "fully_dockerized_benchmark.py"
+    ],
+    deps = [
+        "//src/lib:docker_image",
+        "//src/lib:docker_volume",
+    ],
+    visibility = [ "//visibility:public" ]
```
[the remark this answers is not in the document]
abaptiste: ack

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -0,0 +1,68 @@
...
+options common to all execution methods
+"""
+import os
+import logging
+
+import src.lib.docker_image as docker_image
+import src.lib.docker_volume as docker_volume
+
+log = logging.getLogger(__name__)
+"""
```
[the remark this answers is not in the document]
abaptiste: certainly

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Yep, I can take a look and make the functions private where applicable

abaptiste: sure

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: I can investigate and add this later once I address the other comments.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: Sure.  For now it's just one argument.  I will update.   There are other places where `kwargs` is used like this so I'll take a look at those also.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: Right, this is true.  We shouldn't rely on the user of the class to call validate.  I'll make this private and update the PR

abaptiste:
@mum4k This is ready for another pass.  Thanks.

I am still looking into the Type Hints and will chat with you about that in slack.

salvo/src/lib/benchmark/base_benchmark.py
mum4k: (optional) We certainly don't have to add pytype annotations in this PR (and we aren't going to block this PR on this comment).

I did find the pytype annotations very helpful in preventing bugs and shortening the debug cycle on Python code, so if you are interested and inclined towards using them here I will be happy to work with you and help you figure out how to configure the build system to make them work.

Let's address this in a slack conversation as you have suggested.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -0,0 +1,37 @@
...
+import logging
+
+from src.lib.benchmark.base_benchmark import BaseBenchmark
+
+log = logging.getLogger(__name__)
+
+
+class Benchmark(BaseBenchmark):
+
+  def __init__(self, job_control, benchmark_name, **kwargs):
```
mum4k, line 17: Looks like the comment about `kwargs` was only partially addressed. We do have exact arguments now, but we kept the `kwargs` also.

Does this mean we don't know the complete list of arguments here? And if that is the case, can we discuss this and figure out an architecture where the full list of arguments passed through the layers is known?

If we do know the complete list of arguments, can we remove the `kwargs`?

As you have pointed out - this applies to a few locations in this PR.

mum4k, line 17: Note, an initializer of a public class is also considered public and should have a docstring. When updating the other docstrings to the format recommended by the style guide, let's also add a docstring here and include an `Args:` section where we describe the arguments the initializer takes.

mum4k (changes requested):
Thanks for iterating, looks like a couple of the review comments need a bit more attention, or need to be applied in more locations. I have left comments in location where this applies.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: I'll remove it here since it's unused.  The other places pass kwargs to things like `subprocess.check_output` and the method to run a docker image.  In those cases kwargs seems much more compact than having a long list of parameters and values.

abaptiste:
@mum4k Addressed your comments.  Please review when you have a chance.  Thanks!

salvo/src/lib/job_control_loader.py
```diff
@@ -13,9 +13,17 @@
...
+  """Load a disk file as JSON.
+
+  This function reads the specified filename and parses the contents
+  as JSON.
+
+  Args:
+      filename: The file whose contents are to be read as JSON data
+
+  Returns:
+      A dictionary representation of the file contents
```
mum4k, line 25: When describing a complex container, we should be more explicit about its structure and the inner object types to help the reader / user.

Example for this dictionary: {'John Doe': 345234}

Returns:
  A dict of strings to integers, the keys are names of people and the
  values are their phone numbers. Example:
    {'John Doe': 345234}

Including an example section is optional and I would only include it for more complex dictionaries.

Note - this is another requirement that goes away if we choose to use the pytype annotations.

This comment applies to a few locations.

mum4k (approved):
Thank you @abaptiste, this looks good. Including the last set of comments.

@htuch please review and merge once the remaining comments are resolved.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def get_docker_volumes(output_dir, test_dir=''):
+  """Build the volume structure needed to run a container.
+
+  Build the json specifying the volume configuration needed for running the container
+
+  Args:
+      output_dir: The directory containing the artifacts of the benchmark
+      test_dir: If specified, points to the location of user supplied tests
```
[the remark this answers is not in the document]
abaptiste: updated

salvo/src/lib/constants.py
```diff
@@ -0,0 +1,6 @@
+"""
+Constants module
+"""
+
+DOCKER_SOCKET_PATH = '/var/run/docker.sock'
```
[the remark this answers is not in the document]
abaptiste: Updated

salvo/src/lib/job_control_loader.py
abaptiste: Can we revisit this point in a later PR. I have pytypes working locally and can add annotations to this module:

```
$ grep ^"def _load_json_doc" src/lib/job_control_loader.py
def _load_json_doc(filename: str) -> JobControl:
$ pytype src -P bazel-bin:.
Computing dependencies
Analyzing 25 sources with 6 local dependencies
ninja: Entering directory `/home/ubuntu/code/salvo/.pytype'
[31/31] check src.lib.service.test_s3_upload_event
Leaving directory '/home/ubuntu/code/salvo/.pytype'
Success: no errors found

```

abaptiste: ptypes are coming.  I was able to get this going locally.

mum4k: Great news! Note that even with pytype, we should still say the meaning of the keys and values. Pytype removes the need to specifically mention the object types.

Something like:

```
Returns:
  A dictionary, keys are the file names and values are the file content.
```

abaptiste:
@htuch This is ready for your review.  Thanks!

htuch:
@envoyproxy/senior-maintainers agreed to make @mum4k a maintainer of this repo (based on Nighthawk history), so I think we can merge once he is satisfied. Thanks.

mum4k:
Thank you @htuch, this will decrease the number of review rounds on each PR and help us maintain a good velocity while developing Salvo. We will certainly keep you in the loop for any major architecture decisions or direction changes.

@abaptiste if you can, please address the remaining comment about documenting the meaning of keys and values of some of the dictionaries. We can also do that in a later PR if you want. I will give this PR one more glance and merge later tonight.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+log = logging.getLogger(__name__)
+
+
+def get_docker_volumes(output_dir, test_dir=''):
+  """Build the volume structure needed to run a container.
+
+  Build the json specifying the volume configuration needed for running the container
+
+  Args:
+      output_dir: The directory containing the artifacts of the benchmark
```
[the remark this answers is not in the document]
abaptiste: I would prefer to go with pytype annotations.  I'm not certain whether the action here is to add the object types to the documentation in addition to the annotations.

I do have the annotations working and can include those in the next PR.

mum4k (approved):

PR #75, opened 2020-11-20
Fully dockerized flow - add pytype annotations

abaptiste:
This PR contains two changes
- Add and update dependencies to run the `pytype` checker on the source
- Add the `pytype` annotations to the existing skeleton code.

Note that we are not running any `pytype` checks just yet.  We'll integrate that into the build once our stubs are upgrade to return real types.

Additionally some doc strings were updated to be more descriptive and add more depth. We will iterate more on the doc strings as needed.

We still use `kwargs` in some places since it's more compact than a list of arguments.  We have not added annotations in all places since the validation seems somewhat inconsistent.   We can update these annotations as more functions are implemented, and also eliminate kwargs once the argument list remain stable.

cc: @mum4k, @landesherr


salvo/src/lib/docker_volume.py
```diff
@@ -55,8 +55,11 @@ def generate_volume_config(output_dir, test_dir=''):
   volume_json = {}
   try:
     volume_json = json.loads(MessageToJson(volume_cfg))
-  except Error as serialize_error:
-    log.exception(f"Could not build volume json object: {serialize_error}")
+  except json.decoder.JSONDecodeError as decode_error:
+    log.error(f"Could not build volume json object: {decode_error}")
+    raise
```
mum4k, line 60: (pre-existing) If a function directly raises an exception, we should document it in the "Raises:" section of its docstring.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#24-exceptions

salvo/salvo.py
```diff
@@ -13,14 +13,22 @@
 log = logging.getLogger()
 
 
-def setup_logging(loglevel=logging.DEBUG):
-  """Basic logging configuration """
+def setup_logging(loglevel: int=logging.DEBUG) -> None:
```
mum4k, line 16: (pre-existing) Since we are here and are improving the docstring, can we also add an "Args:" section and describe the loglevel argument? As discussed previously, we don't have to specify the type since we do have pytype, but we could say something about its meaning in the context of this function and Salvo as a whole.

salvo/tools/typecheck.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
mum4k, line 1: (fyi) Sorry to see that pytype_binary() and pytype_library() bazel targets aren't available in OSS yet.

https://github.com/bazelbuild/bazel/issues/4948

Thank you for figuring out how to perform type checks without them. Do we want to file an issue talking about adding this script to the CI so that we can track it?

mum4k (changes requested):
This is great, thank you @abaptiste.

salvo/src/lib/docker_volume.py
abaptiste: Updated

salvo/salvo.py
abaptiste: Updated.

salvo/tools/typecheck.sh
abaptiste: Certainly.  I added https://github.com/envoyproxy/envoy-perf/issues/76 to track this.

abaptiste:
@mum4k This is ready for another pass from you.  Please let me know if there are any other required changes.

mum4k (approved):
Looks good, thank you.

Issue #76, opened 2020-11-23
[Salvo] Add pytype script to the CI checks


abaptiste:
This is a tracking issue so that we do not miss adding the pytype checks to the CI build scripts.

We are not currently enforcing the type checks since most of the implementation is a skeleton.

abaptiste:
Closing this since the type checker is added to CI.

PR #79, opened 2020-12-01
[Salvo] Add method implementation and populate unit tests

abaptiste:
In this PR we add the implementation for the skeleton methods, and the accompanying unit tests.

Doc strings were adjusted based on pylint warnings.  The pytype checks were added to the CI script.  This should resolve https://github.com/envoyproxy/envoy-perf/issues/76.

cc: @mum4k, @landesherr

### DIFF
--- a/salvo/src/lib/benchmark/base_benchmark.py
+++ b/salvo/src/lib/benchmark/base_benchmark.py
@@ -15,10 +15,11 @@ from api.source_pb2 import SourceRepository
 log = logging.getLogger(__name__)
 
 
-def get_docker_volumes(output_dir: str, test_dir: str='') -> dict:
+def get_docker_volumes(output_dir: str, test_dir: str = '') -> dict:
   """Build the volume structure needed to run a container.
 
-  Build the json specifying the volume configuration needed for running the container
+  Build the json specifying the volume configuration needed for running the
+  container.
 
   Args:
       output_dir: The directory containing the artifacts of the benchmark
@@ -40,7 +41,19 @@ class BaseBenchmark(object):
           of benchmark artifacts
         benchmark_name: The name of the benchmark to execute
     """
-    pass
+    if job_control is None:
+      raise Exception("No control object received")
+
+    self._docker_image = docker_image.DockerImage()
+    self._control = job_control
+    self._benchmark_name = benchmark_name
+
+    self._mode_remote = self._control.remote
+    self._build_envoy = False
+    self._build_nighthawk = False
+
+    log.debug("Running benchmark: %s %s", "Remote" if self._mode_remote \
+      else "Local", self._benchmark_name)
 
   def is_remote(self) -> bool:
     """Return a boolean indicating whether the test is to be executed
@@ -49,7 +62,7 @@ class BaseBenchmark(object):
     Returns:
         Whether or not the benchmark runs locally or in a remote service
     """
-    pass
+    return self._mode_remote
 
   def get_images(self) -> DockerImages:
     """Return the images object from the control object.
@@ -57,7 +70,7 @@ class BaseBenchmark(object):
     Returns:
         The images objects specified in the control object
     """
-    pass
+    return self._control.images
 
   def get_source(self) -> List[SourceRepository]:
     """Return the source object defining locations from where
@@ -66,9 +79,9 @@ class BaseBenchmark(object):
     Returns:
         The source objects specified in the control object
     """
-    pass
+    return self._control.source
 
-  def run_image(self, image_name : str, **kwargs) -> bytearray:
+  def run_image(self, image_name: str, **kwargs) -> bytearray:
     """Run the specified docker image with the supplied keyword arguments.
 
     Args:
@@ -78,20 +91,54 @@ class BaseBenchmark(object):
           here https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run
 
     Returns:
-        A bytearray containing the output produced from executing the specified container
+        A bytearray containing the output produced from executing the specified
+        container
     """
-    pass
+    return self._docker_image.run_image(image_name, **kwargs)
 
   def pull_images(self) -> List[str]:
     """Retrieve all images necessary for the benchmark.
 
-    Retrieve the NightHawk and Envoy images defined in the control object. 
+    Retrieve the NightHawk and Envoy images defined in the control
+    object.
     """
-    pass
+    retrieved_images = []
+    images = self.get_images()
+
+    for image in [
+        images.nighthawk_benchmark_image,
+        images.nighthawk_binary_image,
+        images.envoy_image
+    ]:
+      # If the image name is not defined, we will have an empty string.
+      # For unit testing we'll keep this behavior. For true usage, we
+      # should raise an exception when the benchmark class performs its
+      # validation
+      if image:
+        i = self._docker_image.pull_image(image)
+        log.debug(f"Retrieved image: {i} for {image}")
+        if i is None:
+          return []
+        retrieved_images.append(i)
+
+    return retrieved_images
 
   def set_environment_vars(self) -> None:
     """Build the environment variable map used to launch an image.
 
-    Set the Envoy IP test versions and any other environment variables needed by the test
+    Set the Envoy IP test versions and any other environment variables needed
+    by the test
     """
-    pass
+    environment = self._control.environment
+    if not environment:
+      raise Exception("No environment variables are specified")
+
+    if environment.test_version == environment.IPV_UNSPECIFIED:
+      raise Exception("No IP version is specified for the benchmark")
+    elif environment.test_version == environment.IPV_V4ONLY:
+      os.environ['ENVOY_IP_TEST_VERSIONS'] = 'v4only'
+    elif environment.test_version == environment.IPV_V6ONLY:
+      os.environ['ENVOY_IP_TEST_VERSIONS'] = 'v6only'
+
+    for key, value in environment.variables.items():
+      os.environ[key] = value
--- a/salvo/src/lib/benchmark/base_benchmark.py
+++ b/salvo/src/lib/benchmark/base_benchmark.py
@@ -6,11 +6,11 @@ import os
 import logging
 from typing import List
 
-from src.lib.docker_image import (DockerImage, DockerRunParameters)
+import src.lib.docker_image  as docker_image
 import src.lib.docker_volume as docker_volume
-from api.control_pb2 import JobControl
-from api.image_pb2 import DockerImages
-from api.source_pb2 import SourceRepository
+import api.control_pb2 as proto_control
+import api.image_pb2 as proto_image
+import api.source_pb2 as proto_source
 
 log = logging.getLogger(__name__)
 
@@ -33,7 +33,7 @@ def get_docker_volumes(output_dir: str, test_dir: str = '') -> dict:
 class BaseBenchmark(object):
   """Base Benchmark class with common functions for all invocations."""
 
-  def __init__(self, job_control: JobControl, benchmark_name: str) -> None:
+  def __init__(self, job_control: proto_control.JobControl, benchmark_name: str) -> None:
     """Initialize the Base Benchmark class.
 
     Args:
@@ -47,7 +47,7 @@ class BaseBenchmark(object):
     if job_control is None:
       raise Exception("No control object received")
 
-    self._docker_image = DockerImage()
+    self._docker_image = docker_image.DockerImage()
     self._control = job_control
     self._benchmark_name = benchmark_name
 
@@ -67,7 +67,7 @@ class BaseBenchmark(object):
     """
     return self._mode_remote
 
-  def get_images(self) -> DockerImages:
+  def get_images(self) -> proto_image.DockerImages:
     """Return the images object from the control object.
 
     Returns:
@@ -75,7 +75,7 @@ class BaseBenchmark(object):
     """
     return self._control.images
 
-  def get_source(self) -> List[SourceRepository]:
+  def get_source(self) -> List[proto_source.SourceRepository]:
     """Return the source object defining locations from where
        NightHawk or Envoy can be built.
 
@@ -84,12 +84,14 @@ class BaseBenchmark(object):
     """
     return self._control.source
 
-  def run_image(self, image_name: str, run_parameters: DockerRunParameters) -> bytearray:
+  def run_image(self, image_name: str,
+                run_parameters: docker_image.DockerRunParameters) -> bytearray:
     """Run the specified docker image with the supplied keyword arguments.
 
     Args:
         image_name: The docker image to be executed
-        run_paramters: a namedtuple of parameters supplied to an image for execution
+        run_paramters: a namedtuple of parameters supplied to an image for
+          execution
 
     Returns:
         A bytearray containing the output produced from executing the specified
@@ -132,15 +134,16 @@ class BaseBenchmark(object):
 
     return retrieved_images
 
-  def set_environment_vars(self) -> None:
+  def _set_environment_vars(self) -> None:
     """Build the environment variable map used to launch an image.
 
     Set the Envoy IP test versions and any other environment variables needed
-    by the test
+    by the test. This method is called before we execute the docker image so
+    that the image has all variables it needs for a given benchmark.
     """
+    self._clear_environment_vars()
+
     environment = self._control.environment
-    if not environment:
-      raise Exception("No environment variables are specified")
 
     if environment.test_version == environment.IPV_UNSPECIFIED:
       raise Exception("No IP version is specified for the benchmark")
@@ -149,5 +152,25 @@ class BaseBenchmark(object):
     elif environment.test_version == environment.IPV_V6ONLY:
       os.environ['ENVOY_IP_TEST_VERSIONS'] = 'v6only'
 
+    if environment.envoy_path:
+      os.environ['ENVOY_PATH'] = environment.envoy_path
+
     for key, value in environment.variables.items():
       os.environ[key] = value
+
+  def _clear_environment_vars(self) -> None:
+    """Clear any environment variables in the job control document
+       so that we do not influence additionally executing tests.
+    """
+    environment = self._control.environment
+
+    # Check that the key exists before deleting it to prevent KeyErrors
+    if 'ENVOY_IP_TEST_VERSIONS' in os.environ:
+      del os.environ['ENVOY_IP_TEST_VERSIONS']
+
+    if 'ENVOY_PATH' in os.environ:
+      del os.environ['ENVOY_PATH']
+
+    for key, _ in environment.variables.items():
+      if key in os.environ:
+        del os.environ[key]
--- a/salvo/src/lib/benchmark/base_benchmark.py
+++ b/salvo/src/lib/benchmark/base_benchmark.py
@@ -6,7 +6,7 @@ import os
 import logging
 from typing import List
 
-import src.lib.docker_image as docker_image
+from src.lib.docker_image import (DockerImage, DockerRunParameters)
 import src.lib.docker_volume as docker_volume
 from api.control_pb2 import JobControl
 from api.image_pb2 import DockerImages
@@ -40,11 +40,14 @@ class BaseBenchmark(object):
         job_control: The protobuf object containing the parameters and locations
           of benchmark artifacts
         benchmark_name: The name of the benchmark to execute
+
+    Raises:
+        an Exception if no job control object is specified
     """
     if job_control is None:
       raise Exception("No control object received")
 
-    self._docker_image = docker_image.DockerImage()
+    self._docker_image = DockerImage()
     self._control = job_control
     self._benchmark_name = benchmark_name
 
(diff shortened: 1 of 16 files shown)


salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -78,20 +91,54 @@ def run_image(self, image_name : str, **kwargs) -> bytearray:
...
+        container
     """
-    pass
+    return self._docker_image.run_image(image_name, **kwargs)
 
   def pull_images(self) -> List[str]:
     """Retrieve all images necessary for the benchmark.
 
-    Retrieve the NightHawk and Envoy images defined in the control object. 
+    Retrieve the NightHawk and Envoy images defined in the control
```
mum4k, line 102: Can we add a "Returns:" section here explaining the return value? It also looks like there is a specific case when we return an empty list. Can we document that too in the "Returns" section?

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -78,20 +91,54 @@ def run_image(self, image_name : str, **kwargs) -> bytearray:
...
+    ]:
+      # If the image name is not defined, we will have an empty string.
+      # For unit testing we'll keep this behavior. For true usage, we
+      # should raise an exception when the benchmark class performs its
+      # validation
+      if image:
+        i = self._docker_image.pull_image(image)
+        log.debug(f"Retrieved image: {i} for {image}")
+        if i is None:
+          return []
```
mum4k, line 121: Looks like we bail out after the first failure - just making sure that is intentional.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -78,20 +91,54 @@ def run_image(self, image_name : str, **kwargs) -> bytearray:
...
+        retrieved_images.append(i)
+
+    return retrieved_images
 
   def set_environment_vars(self) -> None:
     """Build the environment variable map used to launch an image.
 
-    Set the Envoy IP test versions and any other environment variables needed by the test
+    Set the Envoy IP test versions and any other environment variables needed
```
mum4k, line 129: (pre-existing) Does it make sense for this method to be public? Are there any run modes in which we wouldn't want to setup the environment?

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -78,20 +91,54 @@ def run_image(self, image_name : str, **kwargs) -> bytearray:
...
+    by the test
     """
-    pass
+    environment = self._control.environment
```
mum4k, line 132: Can we add a unit test for the `set_environment_vars` method?

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -7,7 +7,7 @@
 
 import logging
 
-from src.lib.benchmark.base_benchmark import BaseBenchmark
+from src.lib.benchmark.base_benchmark import (BaseBenchmark, get_docker_volumes)
```
mum4k, line 10: Sorry I must have missed this before, can we change the import to one that imports the module rather than classes and methods?

```python
from src.lib.benchmark import base_benchmark

# And then in the code:
base_benchmark.BaseBenchmark()
base_benchmark.get_docker_volumes()
```

The style guide recommends we always import modules to make it easier for readers to identify the source of each symbol.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#22-imports

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -24,25 +24,132 @@ def __init__(self, job_control: JobControl, benchmark_name: str) -> None:
...
         None
+
+    Raises:
+        an Exception if no source definitions allow us to build
+          missing docker images.
     """
-    pass
+    source = self.get_source()
+    if not source:
+      raise Exception("No source configuration specified")
```
mum4k, line 67: Typically it is a good practice to define our own exception for each module. Something like:

```python
class Error(Exception):
  """Raised on errors in this module."""
```

This way the callers of functions can choose to refer to our exception in `try->except` statements. Without this, they can only choose to catch `Exception` itself which covers pretty much any error in Python.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -24,25 +24,132 @@ def __init__(self, job_control: JobControl, benchmark_name: str) -> None:
...
+          and source_def.identity == source_def.SRCID_NIGHTHAWK and \
+          (source_def.source_path or source_def.source_url):
+        can_build_nighthawk = True
+
+      if (not images.envoy_image and not can_build_envoy) or \
+          (not images.nighthawk_benchmark_image or not images.nighthawk_binary_image) \
+              and not can_build_nighthawk:
+        # If the Envoy image is specified, then the validation failed for
+        # NightHawk and vice versa
+        msg = "No source specified to build unspecified {image} image".format(
```
mum4k, line 95: nit: Can we spell out the full name for the variable, i.e. `message` instead of just `msg`? The style guide advises against abbreviations:

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#316-naming

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -24,25 +24,132 @@ def __init__(self, job_control: JobControl, benchmark_name: str) -> None:
...
+        'TMPDIR': output_dir
+    }
+    log.debug(f"Using environment: {image_vars}")
+
+    volumes = get_docker_volumes(output_dir, test_dir)
+    log.debug(f"Using Volumes: {volumes}")
+    self.set_environment_vars()
+
+    kwargs = {}
+    kwargs['environment'] = image_vars
```
mum4k, line 137: This looks like a good time to convert the `kwargs` to a more structured and better documented parameter to `run_image()`.

We can use something very similar to `kwargs`, but with a fixed and documented structure a `namedtuple`. The advantage is that the structure becomes documented and is easier to understand. There are two ways to define and document a namedtuple, feel free to choose based on your preference.

```python
import collections

class LocalRunParameters(collections.namedtuple("LocalRunParameters", "command volumes network_mode tty")):
  """Describes parameters of a local execution.

  Attributes:
    command: A string, ...
    volumes: ...
    network_mode: ...
    tty: ...
  """

or this:

# Describes parameters of an animal.
LocalRunParameters = collections.namedtuple('LocalRunParameters', [
    'command',  # A string, ...
    'volumes',    # ...
    'network_mode',      # ...
    'tty', # ...
])
```

The problem with plain `kwargs` is that the reader of the code has absolutely no idea what parameters are supported or required, nor what is their meaning.

salvo/src/lib/benchmark/test_fully_dockerized_benchmark.py
```diff
@@ -2,25 +2,81 @@
...
 def test_images_only_config():
-  """Test benchmark validation logic."""
-  pass
+  """Test benchmark validation logic.
 
+  Validate that we attempt to run the specified docker image
+  when all parameters are present.
+  """
+  # create a valid configuration defining images only for benchmark
+  job_control = JobControl()
```
mum4k, line 19: (optional) You may or may not like the kwargs based proto API which allows slightly less repetitive constructs like:

```python
job_control = JobControl(
    remote=False,
    scavenging_benchmark=True,
)
```

salvo/src/lib/benchmark/test_fully_dockerized_benchmark.py
```diff
@@ -2,25 +2,81 @@
...
+  docker_images.reuse_nh_images = True
+  docker_images.nighthawk_benchmark_image = \
+      "envoyproxy/nighthawk-benchmark-dev:test_images_only_config"
+  docker_images.nighthawk_binary_image = \
+      "envoyproxy/nighthawk-dev:test_images_only_config"
+  docker_images.envoy_image = \
+      "envoyproxy/envoy-dev:f61b096f6a2dd3a9c74b9a9369a6ea398dbe1f0f"
+
+  env = job_control.environment
+  env.variables["TMP_DIR"] = "/home/ubuntu/nighthawk_output"
```
mum4k, line 33: (optional) Are we planning to make the output directory value a configurable option?

salvo/src/lib/benchmark/test_fully_dockerized_benchmark.py
```diff
@@ -29,34 +85,166 @@ def test_source_to_build_envoy():
   Validate that sources are defined that enable us to build the Envoy image
   We do not expect the validation logic to throw an exception
   """
-  pass
+  # create a valid configuration with a missing Envoy image
+  job_control = JobControl()
```
mum4k, line 89: This test setup code is fairly verbose and looks repetitive. Unless there are significant differences between these test cases, we should consider extracting it to a helper function. As it stands, it distracts the reader from the core concepts of the test cases.

We could have a function that returns the protos setup for the general case and then simply adjust what we need in each test case. That should also make it easier for the reader to spot the "differences" between the test cases which are hard to identify in the current version.

mum4k (changes requested):

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -40,7 +41,19 @@ def __init__(self, job_control: JobControl, benchmark_name: str) -> None:
           of benchmark artifacts
         benchmark_name: The name of the benchmark to execute
     """
-    pass
+    if job_control is None:
+      raise Exception("No control object received")
```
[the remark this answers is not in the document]
abaptiste: Updated

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -78,20 +91,54 @@ def run_image(self, image_name : str, **kwargs) -> bytearray:
...
+        images.nighthawk_benchmark_image,
+        images.nighthawk_binary_image,
+        images.envoy_image
+    ]:
+      # If the image name is not defined, we will have an empty string.
+      # For unit testing we'll keep this behavior. For true usage, we
+      # should raise an exception when the benchmark class performs its
+      # validation
+      if image:
+        i = self._docker_image.pull_image(image)
```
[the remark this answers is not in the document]
abaptiste: Updated.  Used "retrieved_image"

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Updated

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Yes.  We aren't using this method yet, outside of tests, so we can adjust this behavior later if necessary.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Two of the benchmark modes need to set `ENVOY_IP_TEST_VERSIONS`  and others would use this method to set other specified environment variables inherent to that benchmark (eg, `HEAPPROFILE`, `HEAPCHECK`, `ENVOY_PATH`).

Do you feel it should be private?

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: Certainly, this isn't javascript after all :)  Will update.

salvo/src/lib/benchmark/test_fully_dockerized_benchmark.py
abaptiste: `TMP_DIR` is set to the output directory specified in the job control object.  It is configurable.

salvo/src/lib/benchmark/test_fully_dockerized_benchmark.py
abaptiste: Did not know you could do that! Let me work on this.

salvo/src/lib/benchmark/test_fully_dockerized_benchmark.py
abaptiste: Yep.  Will update.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: I agree about `kwargs`. Let me look into using this method

abaptiste:

salvo/src/lib/benchmark/base_benchmark.py
mum4k: Can we also add a sentence about the case when it returns an empty bytearray? Something like:

Returns an empty bytearray if any of the images cannot be retrieved.

This brings another question (which also can be addressed later) - should it return an empty value or would raising an exception be a better choice? Returning empty doesn't really tell the caller that something failed so that error might go unnoticed and the program will continue executing failing in an unexpected way without a good error message.

salvo/src/lib/benchmark/base_benchmark.py
mum4k: That does sound like we have two different run modes. The reason I feel something is wrong with this method is that it is exposed on the API, but takes no arguments or parameters. So it feels like we only get a choice of calling it or not calling it. Looks like all the run modes you described will be calling it.

Since some of the benchmark modes actually need it to set other variables, how do they indicate to this method that other variables should be set? (since this method takes no arguments). Is it by setting `self._control.environment`? Where or what sets the `self._control.environment` attribute? It doesn't seem to be initialized anywhere.

If they do control the variables to be set by setting the `self._control.environment`, I would assume there will be a method that does that. Once we have such a method, maybe that can be combined with this one and there will no longer be a need for this as is? Not sure if I am explaining my confusion correctly, feel free to reach out on slack if I am not.

salvo/src/lib/benchmark/base_benchmark.py
mum4k: This comment doesn't seem to be addressed yet.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
mum4k: This comment doesn't seem to be addressed yet. In addition to this location, we should address the same in `salvo/src/lib/test_docker_image.py` also which was changed in one of the recent commits.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
mum4k: This comment doesn't seem to be addressed yet.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
mum4k: That looks great, can we also add a top level comment above all the locations where we defined a `namedtuple`, since these are public classes in essence?

Example for `DockerRunParameters`
```python
# Provides parameters used when running benchmark in docker.
DockerRunParameters = collections.namedtuple("DockerRunParameters", [
    'environment',  # a dict with environment variable to set in the container
    'command',      # a string containing the command to execute
    'volumes',      # a dict with the volumes mounted in the container
    'network_mode', # a string that specifies the network stack used
    'tty'           # a boolean indicating if a pseudo-tty is allocated
])
```

(optional) Consider adding `pytype` annotations to the namedtuple, see `typing.NamedTuple` which should work just like collections.namedtuple:

```python
Point = typing.NamedTuple("Point", [('x', int), ('y', int)])
```

mum4k:
Thanks @abaptiste, made the second pass and marked all addressed comments as resolved. There are a few left unaddressed in the history above and I have added a few notes on others. Please take a look and let me know.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Raising an Exception might be better here.  I think it makes the logic a bit easier to handle in the event one of the images is unavailable.

I'll update that and the docs.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: From the doc [here](https://github.com/envoyproxy/nighthawk/blob/master/benchmarks/README.md), there are three execution modes.

This method operates on the values set in `self._control.environment` by the user.  

```
$ cat local_fully_dockerized_sources_only.json
{
  "remote": false,
  "dockerizedBenchmark": true,
  "images": {
    "reuseNhImages": true,
    "nighthawkBenchmarkImage": "envoyproxy/nighthawk-benchmark-dev:latest",
    "nighthawkBinaryImage": "envoyproxy/nighthawk-dev:latest"
  },
  "source": [{
    "identity": "SRCID_ENVOY",
    "source_url": "https://github.com/envoyproxy/envoy.git",
    "commit_hash": "d9af0e985a2fef2c921641b12f4786a658ad0923"
  }],
  "environment": {
    "testVersion": "IPV_V4ONLY",
    "envoyPath": "envoy",
    "outputDir": "/home/ubuntu/nighthawk_output",
    "testDir": "/home/ubuntu/nighthawk_tests"
  }
}
```

We look for the required variables that we must set, and any other free-form variables and values are specified in a map.   We call this method just before running the container, and I anticipate that each mode will be calling this method. 

I can ping you on slack a bit later to discuss.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: I was able to get this going for one of the `NamedTuple` instances.  I'll add a `#TODO` and work on the typing later.  `Dicts` seem to be troublesome

salvo/src/lib/benchmark/base_benchmark.py
mum4k: Thank you for the added details, there is something I don't understand here and talking on Slack may be the best. It is still unclear to me who and how sets the `self._control.environment`. Let's chat tomorrow on Slack, I am sure you will be able to help me identify what I misunderstood.

mum4k:
Thank you for the recent updates @abaptiste. I have marked few more conversations as resolved. There are still three unaddressed comments, just making sure you can see them, because Github actually hides them for me under a "3 hidden conversations, Load more..." button. Just making sure you know about them and are planning to address them.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Missed this one from the main conversation tab.   I will add a test.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: Yep.  I can change this.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: I've created https://github.com/envoyproxy/envoy-perf/issues/82 to address this.

I do have a note to do this exact thing, however since this PR is getting large, I'd like to defer this to the next one.

mum4k: That is a very good point, let's leave it to a separate PR to keep them short. Thank you for suggesting it.

abaptiste:
@mum4k This is ready for another pass.  Thanks!

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
```diff
@@ -130,11 +131,11 @@ def execute_benchmark(self) -> None:
     }
     log.debug(f"Using environment: {image_vars}")
 
-    volumes = get_docker_volumes(output_dir, test_dir)
+    volumes = base_benchmark.get_docker_volumes(output_dir, test_dir)
     log.debug(f"Using Volumes: {volumes}")
-    self.set_environment_vars()
+    self._set_environment_vars()
```
mum4k, line 136: (optional, for a future PR) This pattern where we need to do something and then clear something up tends to be brittle in Python when written like this. Say we have a pattern like this:

```python
self._set_environment_vars()

some_code

self._clear_environment_vars()

```

We can find two ways how this can get broken and end up not clearing the variables:

1. If **some_code** raises any sort of exception, the execution terminates immediately and `self._clear_environment_vars()` never gets called.
1. If a future code change adds a new return statement, a bug is born. The developer developer adding the new return might easily miss the need to cleanup. An example of such a change:

```python
self._set_environment_vars()

some_code
if some_condition:
  return # Exits without clearing.

self._clear_environment_vars()

```

Python's [data model](https://docs.python.org/3/reference/datamodel.html) has an answer for this problem - the concept of a [context manager](https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers) that can be used in a [with statement](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement), changing our code to this form:

```python
with some_environment_controller:
  some_code
```

To achieve this, we would define a new object specifically responsible for setting and clearing the environment variables and implement methods `__enter__` and `__exit__` on them as described [here](https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers). Essentially the `self._set_environment_vars()` method would become `__enter__` and the `self._clear_environment_vars()` method would become `__exit__`. 

This solves both the problems, because the `__exit__` method is called even if exceptions are raised and it is also impossible to add early returns that would bypass `__exit__` as long as the code continues to utilize the with statement.

mum4k (approved):
Looks good, just one optional comment.

salvo/src/lib/benchmark/fully_dockerized_benchmark.py
abaptiste: Ack.  Will work on this.

PR #80, opened 2020-12-03
Add the ability to specify additional benchmark images

abaptiste:
Extend the `source` and `api` protos so that users can specify an arbitrary number of benchmark images.

The intent is to give the users the ability to have more than two points for benchmarks.  In the current mode,  the specified commit hash and one commit prior is tested.  With this change, the user can manually specify other points for which we test.

cc:  @mum4k, @landesherr


salvo/requirements.txt
```diff
@@ -1,34 +1,2 @@
-appdirs==1.4.4
-attrs==17.4.0
-boto3==1.16.22
-botocore==1.19.22
-certifi==2018.1.18
-chardet==3.0.4
-decorator==4.4.2
-distlib==0.3.1
 docker==2.5.1
```
mum4k, line 1: Did we intend to delete all these entries from `requirements.txt`? The PR description doesn't seem to be mentioning it or the reason for it. If it is not directly relevant to this PR, can we move it to a separate one that needs this change?

mum4k (changes requested):
Looks good, just one question about the change to `requirements.txt` which may have been added to this PR by accident. Feels like it might not belong.

salvo/requirements.txt
abaptiste: I made this change when the initial PR commit failed in the CI pipeline.   I keep seeing conflicts about the version of `requests`, so I trimmed the requirements to the minimum items required and let `pip3` figure out the rest.

I can revert this commit, though the CI failure will continue (and I can reproduce locally).

abaptiste: Update.  Seems that the requirements was lagging behind other branches.  Not sure how that happened.  In any case I updated it so that it doesn't fail anymore.  The contents match the `add_method_definitions` branch/PR, so we may see an easily resolved conflict when things are merged.

mum4k: Thank you.

mum4k (approved):

PR #83, opened 2020-12-08
Move docker modules into their own package.  Use custom Errors

abaptiste:
In this PR: 

1.  We move the docker modules into their own package.
2. We define and use custom Errors in where applicable.
3. We use an Environment Controller for setting and clearing environment variables required for a benchmark.

Pylint errors and warnings are continually addressed hence the formatting changes.

cc: @mum4k, @landesherr


salvo/src/lib/docker/BUILD
```diff
@@ -0,0 +1,39 @@
...
+
+licenses(["notice"])  # Apache 2
+
+package(
+    default_visibility = ["//:__subpackages__"],
+)
+
+py_library(
+    name = "docker_image",
+    data = [
```
mum4k, line 11: Looks like I completely missed this before. I believe this shouldn't be marked as `data =`, but as `srcs =`. There is a difference, `srcs` tells blaze that these are code source files while `data` instructs blaze to package these files into the generated runfiles directory.

https://docs.bazel.build/versions/master/be/python.html#py_library

If you don't mind, let's clean this up in a separate PR.

salvo/src/lib/docker/docker_volume.py
```diff
@@ -0,0 +1,68 @@
+"""
```
mum4k, line 1: I may be misreading this, but it looks like we are leaving the original [docker_image.py](https://github.com/abaptiste/envoy-perf/blob/custom_exceptions/salvo/src/lib/docker_image.py) and [docker_volume.py](https://github.com/abaptiste/envoy-perf/blob/custom_exceptions/salvo/src/lib/docker_volume.py) behind.

Was the intention to move them to the new directory? If yes, do you mind updating this to a git move using the `git mv` command which will improve both the code review and the git history we leave behind?
 The simplest way to do that is:

1. Backup content of these two files in your local client.
1. Revert these new files.
1. `git mv` from the original files into this location
1. replace the file content from backups, commit and push.

Git history should take care of the rest and the PR should help the reviewer see what if any changes were performed.

mum4k (changes requested):
Looks good, a few comments below.

Not actionable for this PR, but requesting a favor for the future. If you don't mind, it would make code reviews much easier if we separate moves and refactors from any code changes. (moving the docker module and adding the new exceptions is combined in this PR). When the PR does just the move, the reviewer has to read much less code. Ideally let's make every PR do just one thing.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -174,3 +191,20 @@ def _clear_environment_vars(self) -> None:
...
+    """Interface that must be implemented to execute a given
+       benchmark.
+
+    Raises:
+      NotImplementedError: if this method is not overridden
+    """
+    raise NotImplementedError("Execute Benchmark must be implemented")
+
+  def __enter__(self):
+    """Method executed when entering a context."""
```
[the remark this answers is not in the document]
abaptiste: Will update.

salvo/src/lib/benchmark/benchmark_errors.py
```diff
@@ -0,0 +1,15 @@
+"""Module defining Exceptions raised from benchmark classes"""
```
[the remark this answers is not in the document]
abaptiste: Ack.  Will update.

salvo/src/lib/docker/BUILD
abaptiste: Sure thing.  Not sure how this one escaped me either.

salvo/src/lib/docker/docker_volume.py
abaptiste: Those should be removed.  I did not notice that.  Will clean up.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -6,11 +6,13 @@
 import logging
 from typing import List
 
-import src.lib.docker_image  as docker_image
-import src.lib.docker_volume as docker_volume
+import src.lib.docker.docker_image as docker_image
```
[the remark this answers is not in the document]
abaptiste: I will keep this in mind

abaptiste:

abaptiste:
@mum4k This is ready for another pass from you. Thanks!

salvo/src/lib/docker/docker_volume.py
mum4k: Thank you, this made it much easier to review.

mum4k (approved):

PR #84, opened 2020-12-09
Use a context manager to stop lingering docker images

abaptiste:
In this change we add a context manager to enumerate and stop any lingering docker images from the benchmark.

We also fixed a couple outstanding issues:
1.  in a BUILD file "data" is correctly replaced by "srcs".
2.  fix import statements in the test (for now.  Other files are forthcoming)

cc: @mum4k, @landesherr

### DIFF
--- a/salvo/src/lib/docker/docker_image.py
+++ b/salvo/src/lib/docker/docker_image.py
@@ -39,12 +39,19 @@ class DockerImage():
 
     Args:
         image_name: The name of the docker image that we are retrieving from
-        dockerhub or another repository
+        dockerhub or another repository. If the image exists locally, we return
+        the corresponding docker image object from the image name
 
     Returns:
         The Image that was pulled
     """
-    return self._client.images.pull(image_name)
+    existing_images = self.list_images()
+    if image_name not in existing_images:
+      image = self._client.images.pull(image_name)
+    else:
+      image = self._client.images.get(image_name)
+
+    return image
 
   def list_images(self) -> List[str]:
     """List all available docker image tags.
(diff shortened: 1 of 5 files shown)


abaptiste:
Tests are failing due to the lack of the docker socket.  Taking a look.

salvo/src/lib/docker/docker_image.py
```diff
@@ -80,5 +87,55 @@ def run_image(self, image_name: str,
           container
     """
 
-    return self._client.containers.run(image_name, stdout=True, stderr=True,
-                                       detach=False, **run_parameters._asdict())
+    output = ''
+    with DockerImageController(self):
```
mum4k, line 91: This is a nice improvement.

(optional) We could take it a step forward. The current setup still allows other devs to "forget" about the `DockerImageController` and just call `self._client.containers.run` directly. I am actually not sure if it will work without, but the current API doesn't help the reader to understand the relation between the `DockerImageController` and the `DockerImage`.

To address this problem, we would refactor the code so that running with the controller is thge only way to execute. This can be done by moving the `run` method onto the controller, so that this code becomes:

```python
with DockerImageController(self) as controller:
  output = controller.run(...)
```

Note that DockerImage can still retain the implementation of the `run` method, but it would be private. This makes the relationship more obvious, because the controller is actually used for something besides its side effects.

Marking as optional. Please let me know if you would like to address it in this PR, or if we should merge as is.

mum4k (approved):

salvo/src/lib/docker/docker_image.py
abaptiste: @mum4k The docker `_client` is private.  `DockerImage` is a wrapper around the "docker" command and python module.  I'm not sure I follow about how to refactor to achieve your suggestion.   The `run_image` method is what users should use to start a docker image.

Let's go ahead and merge this change.  I will think about it some more and perhaps give it a shot in a future PR.

mum4k: Absolutely, proceeding with the merge.

The main point we may want to address (still remains optional though) is that the context manager instance is not actually used within its context which looks a bit surprising.

PR #88, opened 2021-01-08
Add source_tree implementation and tests

abaptiste:
Add the `source_tree` module that abstracts git operations to manipulate and deduce information about a source tree.

Unit tests for the source tree module are included in this diff.

cc: @mum4k , @landesherr

### DIFF
--- a/salvo/src/lib/common/BUILD
+++ b/salvo/src/lib/common/BUILD
@@ -1,0 +1,12 @@
+load("@rules_python//python:defs.bzl", "py_library")
+
+package(
+  default_visibility = ["//:__subpackages__"],
+)
+
+py_library(
+    name = "fileops",
+    data = [
+        "fileops.py",
+    ],
+)
--- a/salvo/src/lib/common/BUILD
+++ b/salvo/src/lib/common/BUILD
@@ -10,3 +10,11 @@ py_library(
         "file_ops.py",
     ],
 )
+py_test(
+    name = "test_file_ops",
+    srcs = ["test_file_ops.py"],
+    srcs_version = "PY3",
+    deps = [
+        ":file_ops",
+    ],
+)
--- a/salvo/src/lib/common/BUILD
+++ b/salvo/src/lib/common/BUILD
@@ -5,8 +5,8 @@ package(
 )
 
 py_library(
-    name = "fileops",
+    name = "file_ops",
     data = [
-        "fileops.py",
+        "file_ops.py",
     ],
 )
--- a/salvo/src/lib/common/file_ops.py
+++ b/salvo/src/lib/common/file_ops.py
@@ -25,8 +25,7 @@ def open_json(path: str, mode: str = 'r') -> dict:
 
 
 def open_yaml(path: str, mode: str = 'r') -> dict:
-  """Open a json file and return its contents as a dictionary
-
+  """Open a yaml file and return its contents as a dictionary 
   Args:
     path: the full path to the YAML file
     mode: the mode with which we open the file.  The mode defaults
--- a/salvo/src/lib/common/fileops.py
+++ b/salvo/src/lib/common/fileops.py
@@ -1,0 +1,68 @@
+"""
+Module to abstract a few common file operations used in the framework
+"""
+import json
+import random
+import shutil
+import yaml
+import glob
+import os
+import tempfile
+
+
+def open_json(path: str, mode: str = 'r') -> dict:
+  """Open a json file and return its contents as a dictionary
+
+  Args:
+    path: the full path to the JSON file
+    mode: the mode with which we open the file.  The mode defaults
+      to read only and is an optional argument.
+  """
+  data = {}
+  with open(path, mode) as json_file:
+    data = json.loads(json_file.read())
+  return data
+
+
+def open_yaml(path: str, mode: str = 'r') -> dict:
+  """Open a json file and return its contents as a dictionary
+
+  Args:
+    path: the full path to the YAML file
+    mode: the mode with which we open the file.  The mode defaults
+      to read only and is an optional argument.
+  """
+  data = {}
+  with open(path, mode) as yaml_file:
+    data = yaml.load(yaml_file)
+  return data
+
+
+def delete_directory(path: str) -> None:
+  """Delete a directory and its contents.
+
+  Args:
+    path: The directory to be deleted
+  """
+  shutil.rmtree(path)
+
+
+def get_random_dir(path: str) -> str:
+  """Get a random named directory.
+
+  The caller must re-create the path received if the value
+  is used as-is.
+
+  Args:
+    path: The location where the temporary directory is to be
+     created
+
+  Returns:
+    The full pathname of the temporary directory.
+  """
+
+  if not os.path.exists(path):
+    os.mkdir(path)
+
+  temp_dir = tempfile.TemporaryDirectory(dir=path)
+  return temp_dir.name
--- a/salvo/src/lib/common/test_file_ops.py
+++ b/salvo/src/lib/common/test_file_ops.py
@@ -13,7 +13,7 @@ from src.lib.common import file_ops
 from unittest import mock
 
 
-_test_json = """
+_TEST_JSON = """
 {
   "key" : "this is my json file",
   "key2" : "there are many like it",
@@ -23,7 +23,7 @@ _test_json = """
 }
 """
 
-_test_yaml = """
+_TEST_YAML = """
 key: this is my json file
 key2: there are many like it
 key3:
@@ -42,7 +42,7 @@ def _validate_test_data(file_data):
 def test_open_json():
   with tempfile.NamedTemporaryFile() as temp_json:
     with open(temp_json.name, 'w') as temp_data:
-      temp_data.write(_test_json)
+      temp_data.write(_TEST_JSON)
 
     json_data = file_ops.open_json(temp_json.name)
     _validate_test_data(json_data)
@@ -50,7 +50,7 @@ def test_open_json():
 def test_open_yaml():
   with tempfile.NamedTemporaryFile() as temp_yaml:
     with open(temp_yaml.name, 'w') as temp_data:
-      temp_data.write(_test_yaml)
+      temp_data.write(_TEST_YAML)
 
     yaml_data = file_ops.open_yaml(temp_yaml.name)
     _validate_test_data(yaml_data)
@@ -58,7 +58,7 @@ def test_open_yaml():
 def test_open_yaml_as_json():
   with tempfile.NamedTemporaryFile() as temp_json:
     with open(temp_json.name, 'w') as temp_data:
-      temp_data.write(_test_yaml)
+      temp_data.write(_TEST_YAML)
 
     with pytest.raises(json.decoder.JSONDecodeError) as decode_error:
       json_data = file_ops.open_json(temp_json.name)
@@ -68,7 +68,7 @@ def test_open_yaml_as_json():
 def test_open_json_as_yaml():
   with tempfile.NamedTemporaryFile() as temp_yaml:
     with open(temp_yaml.name, 'w') as temp_data:
-      temp_data.write(_test_json)
+      temp_data.write(_TEST_JSON)
 
     yaml_data = file_ops.open_yaml(temp_yaml.name)
     _validate_test_data(yaml_data)
--- a/salvo/src/lib/common/test_file_ops.py
+++ b/salvo/src/lib/common/test_file_ops.py
@@ -1,0 +1,89 @@
+
+"""
+Test source_tree operations needed for executing benchmarks
+"""
+import pytest
+
+import os
+import json
+import tempfile
+
+from src.lib.common import file_ops
+
+from unittest import mock
+
+
+_test_json = """
+{
+  "key" : "this is my json file",
+  "key2" : "there are many like it",
+  "key3" : [
+    "but", "this", "one", "is", "mine"
+  ]
+}
+"""
+
+_test_yaml = """
+key: this is my json file
+key2: there are many like it
+key3:
+  - but
+  - this
+  - one
+  - is
+  - mine
+"""
+
+def _validate_test_data(file_data):
+  assert file_data
+  assert all(map(lambda key: key in file_data, ["key", "key2", "key3"]))
+  assert file_data['key3'] == [ 'but', 'this', 'one', 'is', 'mine' ]
+
+def test_open_json():
+  with tempfile.NamedTemporaryFile() as temp_json:
+    with open(temp_json.name, 'w') as temp_data:
+      temp_data.write(_test_json)
+
+    json_data = file_ops.open_json(temp_json.name)
+    _validate_test_data(json_data)
+
+def test_open_yaml():
+  with tempfile.NamedTemporaryFile() as temp_yaml:
+    with open(temp_yaml.name, 'w') as temp_data:
+      temp_data.write(_test_yaml)
+
+    yaml_data = file_ops.open_yaml(temp_yaml.name)
+    _validate_test_data(yaml_data)
+
+def test_open_yaml_as_json():
+  with tempfile.NamedTemporaryFile() as temp_json:
+    with open(temp_json.name, 'w') as temp_data:
+      temp_data.write(_test_yaml)
+
+    with pytest.raises(json.decoder.JSONDecodeError) as decode_error:
+      json_data = file_ops.open_json(temp_json.name)
+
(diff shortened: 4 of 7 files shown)


salvo/src/lib/test_source_tree.py
```diff
@@ -9,83 +9,240 @@
 from src.lib import (cmd_exec, source_tree)
 import api.source_pb2 as proto_source
 
+logging.basicConfig(level=logging.DEBUG)
 log = logging.getLogger(__name__)
 
 def test_source_tree_object():
   """Verify that we throw an exception if not all required data is present."""
-  pass
+  st = source_tree.SourceTree(proto_source.SourceRepository())
```
mum4k, line 17: The Python style guide advises to use descriptive names, even if they are slightly longer. Can we use a more descriptive identifier instead of just `st`? Consider `source` or `tree`.

Applies to multiple locations below.

salvo/src/lib/test_source_tree.py
```diff
@@ -9,83 +9,240 @@
...
 def _generate_source_tree_from_origin(origin):
   """Build a source tree from a remote url."""
-  pass
+  source_repository = proto_source.SourceRepository(
+      source_path='/tmp',
+      source_url=origin
+  )
+  return source_tree.SourceTree(source_repository)
 
 def mock_run_command_side_effect(*args):
```
mum4k, line 106: Since this is a test helper, future maintainers will need to understand it. Can we unwrap the `*args` into an exact list of arguments and document then and their behavior in the `Args` section of the docstring here?

Applies to all such test helpers below.

salvo/src/lib/test_source_tree.py
```diff
@@ -9,83 +9,240 @@
...
-  pass
+  """
+    Verify that we can identify one commit prior to a specified hash.
+    """
+  origin = 'https://github.com/someawesomeproject/repo.git'
+  st = _generate_source_tree_from_origin(origin)
+
+  commit_hash = 'fake_commit_hash_1'
+
+  mock_check_output.side_effect = _check_output_side_effect
```
mum4k, line 243: A question on complexity - do we need to define `_check_output_side_effect()` as a separate function which combines all possible outputs? This makes the reader scroll up and down from the test case to the function and requires the reader to understand the side_effect function.

Can we just set the return value here in the test so the reader has everything in one location? E.g. just set `mock_check_output.return_value`.

Seems like that would allow us to remove some of the helper code and keep test cases self-contained.

Applies to `_check_output_side_effect_fail` as well.

salvo/src/lib/test_source_tree.py
```diff
@@ -9,83 +9,240 @@
...
+    magic_mock.assert_called_once_with("git remote -v", cmd_params)
 
 def test_get_origin_ssh():
   """Verify that we can determine the origin for a local repository.
 
   In this instance the repo was cloned via ssh.
   """
-  pass
+  remote_string = 'origin  git@github.com:username/reponame.git (fetch)'
+  gitcmd = "git remote -v"
```
mum4k, line 60: (nit) The name should be `git_cmd` since these are two words.

salvo/src/lib/test_source_tree.py
```diff
@@ -94,50 +251,321 @@ def _check_output_side_effect_fail(*args):
...
   """Verify that we can list tags from a repository."""
-  pass
+
+  origin = 'https://github.com/someawesomeproject/repo.git'
+  st = _generate_source_tree_from_origin(origin)
+
+  gitcmd = "git tag --list --sort v:refname"
+  git_output = """v1.0.0
+v1.1.0
+v1.2.0
```
mum4k, line 363: Test data are also code in the sense that they need to be understood and maintained with the code changes. Can we minimize this test data (the list of tags) to the bare minimum necessary to test the behavior?

This will decrease the amount of lines we have to maintain here and make it easier to understand the purpose of the test. It isn't clear at the moment if all of these are needed and / or some of them test some special behaviors.

Applies to the remaining two tag lists below as well.

mum4k (changes requested):

salvo/src/lib/common/BUILD
```diff
@@ -0,0 +1,12 @@
+load("@rules_python//python:defs.bzl", "py_library")
+
+package(
+  default_visibility = ["//:__subpackages__"],
+)
+
+py_library(
+    name = "fileops",
```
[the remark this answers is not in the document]
abaptiste: Certainly.  I will update.

salvo/src/lib/common/fileops.py
```diff
@@ -0,0 +1,68 @@
...
+      to read only and is an optional argument.
+  """
+  data = {}
+  with open(path, mode) as json_file:
+    data = json.loads(json_file.read())
+  return data
+
+
+def open_yaml(path: str, mode: str = 'r') -> dict:
+  """Open a json file and return its contents as a dictionary
```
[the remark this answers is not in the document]
abaptiste: Ack!  Fixed.

salvo/src/lib/common/fileops.py
```diff
@@ -0,0 +1,68 @@
...
+
+  Returns:
+    The full pathname of the temporary directory.
+  """
+
+  if not os.path.exists(path):
+    os.mkdir(path)
+
+  temp_dir = tempfile.TemporaryDirectory(dir=path)
+  return temp_dir.name
```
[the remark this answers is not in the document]
abaptiste: Yeah, I agree, though the directory lifespan management is something I need to focus on a bit more down the road.  Let me create an issue for this and revisit it (soon) since this change reverberates through the rest of Salvo.  I'll work on this today(issue https://github.com/envoyproxy/envoy-perf/issues/90) and update in a subsequent PR.

salvo/src/lib/source_tree.py
```diff
@@ -108,15 +123,33 @@ def get_origin(self) -> str:
...
+    log.debug(f"Valid: {valid} for object: {self}")
+
+    origin_url = self._source_url
+
+    cmd = "git remote -v"
+    if not origin_url:
+      cmd_params = cmd_exec.CommandParameters(cwd=self._source_path)
+      output = cmd_exec.run_command(cmd, cmd_params)
+      for line in output.split('\n'):
+        match = re.match(r'^origin\s*([\w:@\.\/-]+)\s\(fetch\)$', line)
```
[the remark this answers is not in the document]
abaptiste: Updated

salvo/src/lib/common/fileops.py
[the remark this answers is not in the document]
abaptiste: I spent the day working on this and had to punt.  It's not going to land in this PR.

salvo/src/lib/source_tree.py
```diff
@@ -40,24 +40,32 @@ def __init__(self, source_repo: proto_source.SourceRepository) -> None:
...
+      home_dir = constants.SALVO_TMP
+
+    self._tempdir = fileops.get_random_dir(home_dir)
 
     self._source_repo = source_repo
-    self._source_url = source_repo.source_url
-    self._branch = source_repo.branch
-    self._commit_hash = source_repo.commit_hash
-    self._source_path = source_repo.source_path
+    self._source_url = self._source_repo.source_url
```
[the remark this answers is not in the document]
abaptiste: Let me work on this.  I had this in a previous implementation but ended up changing it back.  I'll update.

salvo/src/lib/test_source_tree.py
abaptiste: Fixed

salvo/src/lib/test_source_tree.py
abaptiste: Guess I got a little happy witht he mocks.  I've removed `_check_output_side_effect_fail` and used `return_value` in this case.  I also consoliated the side effect functions as much as I could.

salvo/src/lib/test_source_tree.py
abaptiste: Updated

salvo/src/lib/test_source_tree.py
abaptiste: Updated

salvo/src/lib/source_tree.py
```diff
@@ -158,6 +251,33 @@ def get_previous_commit_hash(self, current_commit: str,
     Returns:
       a string with the discovered commit hash
     """
+
```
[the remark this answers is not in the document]
abaptiste: I believe validate gets called at some point in the call tree. I just made a pass to verify.  If there's something I am still missing, please highlight.

salvo/src/lib/source_tree.py
```diff
@@ -170,23 +290,57 @@ def get_revs_behind_parent_branch(self) -> int:
...
+    if not self._source_path:
+      raise SourceTreeError("Source is not checked out on disk")
+
+    cmd = "git status"
+    cmd_params = cmd_exec.CommandParameters(cwd=self._source_path)
+    status_output = cmd_exec.run_command(cmd, cmd_params)
+
+    commit_count = 0
+    ahead = re.compile(r'.*ahead of \'(.*)\' by (\d+) commit[s]')
+    uptodate = re.compile(r'Your branch is up to date with \'(.*)\'')
```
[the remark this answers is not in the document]
abaptiste: Fixed

salvo/src/lib/source_tree.py
```diff
@@ -170,23 +290,57 @@ def get_revs_behind_parent_branch(self) -> int:
...
+
+    if not self._source_path:
+      raise SourceTreeError("Source is not checked out on disk")
+
+    cmd = "git status"
+    cmd_params = cmd_exec.CommandParameters(cwd=self._source_path)
+    status_output = cmd_exec.run_command(cmd, cmd_params)
+
+    commit_count = 0
+    ahead = re.compile(r'.*ahead of \'(.*)\' by (\d+) commit[s]')
```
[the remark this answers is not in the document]
abaptiste: Updated

salvo/src/lib/source_tree.py
```diff
@@ -108,15 +123,33 @@ def get_origin(self) -> str:
...
+    valid = self._validate()
+    log.debug(f"Valid: {valid} for object: {self}")
+
+    origin_url = self._source_url
+
+    cmd = "git remote -v"
+    if not origin_url:
+      cmd_params = cmd_exec.CommandParameters(cwd=self._source_path)
+      output = cmd_exec.run_command(cmd, cmd_params)
+      for line in output.split('\n'):
```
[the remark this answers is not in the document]
abaptiste: I will add a block to raise an exception if the url is still empty.

salvo/src/lib/source_tree.py
```diff
@@ -127,15 +160,69 @@ def pull(self) -> bool:
     Returns:
       a boolean indicating whether the operation was successful
     """
-    return False
+
+    source_name = proto_source.SourceRepository.SourceIdentity.Name(
+        self._source_repo.identity)
+    log.debug(f"Pulling [{source_name}] from origin: [{self._source_url}]")
+    self._validate()
```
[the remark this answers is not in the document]
abaptiste: I can move the validate call, though I don't think that getting the source identity name from the value would affect things.

abaptiste:
@mum4k This is ready for another pass from you, at your leisure.  Thanks!

salvo/src/lib/common/file_ops.py
```diff
@@ -0,0 +1,67 @@
+"""
```
mum4k, line 1: Sorry I must have missed this - can we also add tests for the `file_ops` package?

salvo/src/lib/common/fileops.py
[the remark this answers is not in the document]
mum4k: Good choice punting this and solving it outside of this PR. Thank you for filing the issue.

salvo/src/lib/source_tree.py
```diff
@@ -108,15 +128,40 @@ def get_origin(self) -> str:
...
+      output = cmd_exec.run_command(cmd, cmd_params)
+      for line in output.split('\n'):
+        match = re.match(_GIT_ORIGIN_REGEX, line)
+        if match:
+          origin_url = match.group(1)
+          self._source_repo.source_url = origin_url
+          break
+
+    if not origin_url:
+      raise SourceTreeError(
```
mum4k, line 149: Can we add the `Raises:` section into the docstring now that the method can raise an exception?

salvo/src/lib/source_tree.py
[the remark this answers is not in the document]
mum4k: Understood, this was mostly a question because the unusual placement (compared to the other methods) implied that this has some semantic meaning. Thanks for unifying it.

salvo/src/lib/source_tree.py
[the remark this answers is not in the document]
mum4k: Thanks for clarifying.

salvo/src/lib/test_source_tree.py
```diff
@@ -9,135 +9,428 @@
 from src.lib import (cmd_exec, source_tree)
 import api.source_pb2 as proto_source
 
+logging.basicConfig(level=logging.DEBUG)
 log = logging.getLogger(__name__)
 
+_GIT_TAG_LIST = """
```
mum4k, line 15: Moving the tag list into a global constant may be a regression. It introduces two problems:

1. The test cases are no longer "complete" or "all in one place" for the reader. The reader has to scroll up and down to understand them.
2. Multiple test cases use the same constant which to the reader implies that they all have to use the same constant / tag list. This likely isn't true and may confuse future maintainers.

How would you feel about moving this back into individual lists in the test cases that need this?

salvo/src/lib/test_source_tree.py
```diff
@@ -9,135 +9,428 @@
...
+v1.14.5
```
mum4k, line 16: Thank you for making the list shorter. This is still bit longer than I expected, indicating that I might be missing something about the test cases. Following the [Zero one infinity rule](https://en.wikipedia.org/wiki/Zero_one_infinity_rule), do we need more than one or two tags to test the behavior we are after? Any more than the bare minimum makes it unclear as to what the tests depend on / really need.

mum4k (changes requested):
Thank you for addressing the comments @abaptiste, sending the last batch.

salvo/src/lib/test_source_tree.py
abaptiste: I wanted more than 2 tags in the list to verify we can discover an arbitrary number of commits backwards from a given hash.  Originally the intent was that we would test the hash N and N-1, but it maybe useful to test N-3 etc.  Having more than 2 tags in the list gives us the ability to verify this.

That said, I shortened the lists where applicable.

salvo/src/lib/test_source_tree.py
abaptiste: Moved them back :)

salvo/src/lib/source_tree.py
abaptiste: Certainly.  Updated.

salvo/src/lib/common/file_ops.py
abaptiste: Added a test for file_ops

abaptiste:
@mum4k Sorry for the delay.  I need a clone.  This is ready for another pass from you.

salvo/src/lib/common/test_file_ops.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+import json
+import tempfile
+
+from src.lib.common import file_ops
+
+from unittest import mock
+
+
+_test_json = """
```
mum4k, line 16: (nit) Constants should use CAPS_WITH_UNDER in their names, so this and the following constant should be named:

_TEST_JSON
_TEST_YAML

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#3164-guidelines-derived-from-guidos-recommendations

mum4k (approved):
Looks good, we have one naming nit, please feel free to indicate if you would like to address it in a follow-up PR.

mum4k:
No worries @abaptiste, there is no time pressure from my end. However when you do figure out how to get a clone, do let me know.

salvo/src/lib/common/test_file_ops.py
abaptiste: I'll take care of it here.

abaptiste:
@mum4k I addressed the last naming change.  Please let me know if you see anything else.  Thanks!

mum4k (approved):
Looks good, thank you @abaptiste and have a nice weekend!

Issue #90, opened 2021-01-12
[Salvo] Use a TemporaryDirectory object vs string when generating temporary directories


abaptiste:
In `file_ops.py` return a TemporaryDirectory object instead of a string so that we can leverage the automatic cleanup done when the object goes out of scope.

PR #91, opened 2021-01-25
Use TemporaryDirectory vs string for temp directories

abaptiste:
This change addresses https://github.com/envoyproxy/envoy-perf/issues/90.  When creating temporary directories we use a `TemporaryDirectory` object instead of a string.

cc: @mum4k , @landesherr

### DIFF
--- a/salvo/src/lib/cmd_exec.py
+++ b/salvo/src/lib/cmd_exec.py
@@ -7,6 +7,7 @@ import shlex
 import subprocess
 import typing
 import logging
+import tempfile
 
 log = logging.getLogger(__name__)
 
@@ -35,18 +36,32 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
       command
   """
   output = ''
+  params = parameters._asdict()
+  tmpfile = tempfile.TemporaryFile(
+      mode='w+', dir=params['cwd'], prefix='cmd_output'
+  )
+
   try:
     log.debug(f"Executing command: [{cmd}] with args [{parameters._asdict()}]")
     cmd_array = shlex.split(cmd)
-    output = subprocess.check_output(
-        cmd_array, stderr=subprocess.STDOUT, **parameters._asdict())
+
+    subprocess.check_call(cmd_array, stdout=tmpfile, stderr=tmpfile,
+                          **parameters._asdict())
+
+  except subprocess.CalledProcessError as process_error:
+    log.error(f"Unable to execute [{cmd}]: {process_error}")
+    raise
+
+  finally:
+    tmpfile.flush()
+    tmpfile.seek(0)
+    output = tmpfile.read()
+    tmpfile.close()
 
     if isinstance(output, bytes):
       output = output.decode('utf-8').strip()
 
     log.debug(f"Returning output: [{output}]")
-  except subprocess.CalledProcessError as process_error:
-    log.error(f"Unable to execute [{cmd}]: {process_error}")
-    raise
 
   return output
+
--- a/salvo/src/lib/cmd_exec.py
+++ b/salvo/src/lib/cmd_exec.py
@@ -35,6 +35,11 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
     subprocess.CalledProcessError if there was a failure executing the specified
       command
   """
+
+  # Because the stdout/stderr from nighthawk can be large, we redirect it to
+  # a temporary file and re-read the output to return to the caller.  This
+  # method also appears to capture output more consistently in the event of a
+  # failed command execution.
   output = ''
   params = parameters._asdict()
   tmpfile = tempfile.TemporaryFile(
--- a/salvo/src/lib/source_tree.py
+++ b/salvo/src/lib/source_tree.py
@@ -1,7 +1,9 @@
 """Manage a source location on disk"""
 import re
+import glob
 import logging
 import os
+import shutil
 import subprocess
 from typing import List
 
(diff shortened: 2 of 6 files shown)


salvo/src/lib/cmd_exec.py
```diff
@@ -35,18 +36,32 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
...
   try:
     log.debug(f"Executing command: [{cmd}] with args [{parameters._asdict()}]")
     cmd_array = shlex.split(cmd)
-    output = subprocess.check_output(
-        cmd_array, stderr=subprocess.STDOUT, **parameters._asdict())
 
-    if isinstance(output, bytes):
-      output = output.decode('utf-8').strip()
+    subprocess.check_call(cmd_array, stdout=tmpfile, stderr=tmpfile,
+                          **parameters._asdict())
```
mum4k, line 49: (nit) Can we use `params` here?

**params that is.

salvo/src/lib/source_tree.py
```diff
@@ -164,8 +166,31 @@ def get_source_directory(self) -> str:
     """
     self._validate()
 
-    return  self._source_repo.source_path \
-                 if self._source_repo.source_path else self._tempdir
+    return self._build_dir.name
+
+  def copy_source_directory(self) -> bool:
```
mum4k, line 171: Are these changes also relevant to #90? If not, can we move them to a separate PR to keep this one focused?

salvo/src/lib/source_tree.py
```diff
@@ -54,7 +56,11 @@ def __init__(self, source_repo: proto_source.SourceRepository) -> None:
     if not home_dir.startswith(constants.SALVO_TMP):
       home_dir = constants.SALVO_TMP
 
-    self._tempdir = file_ops.get_random_dir(home_dir)
+    # This is the destination directory where the source is copied or cloned
+    # If the source_repo has a path, we copy from that path into the _build_dir.
+    # If the source_repo has a url, we clone from that url into the _build_dir.
+    self._build_dir = file_ops.get_random_dir(home_dir)
```
mum4k, line 62: Since we are storing this in an attribute, we aren't really addressing #90, since nothing will destroy the temporary directory after we use it like this. However since this is the build directory we are producing, do we actually want to destroy it after the execution? Or do we want to keep it around for debugging?

If we do want to destroy it, we should ensure that the `__exit__()` method of the created temp directory gets called at some point.

salvo/src/lib/cmd_exec.py
```diff
@@ -35,18 +36,32 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
...
 
-    log.debug(f"Returning output: [{output}]")
   except subprocess.CalledProcessError as process_error:
     log.error(f"Unable to execute [{cmd}]: {process_error}")
     raise
 
+  finally:
+    tmpfile.flush()
```
mum4k, line 56: Is using the temp instead of a string here an improvement? Looks like we have to do some number of operations to get this output now (flush, seek, read). What do we gain by switching this into a file instead of a string. Is this relevant to #90 which was talking about automatic cleanup? Looks like the original solution was already performing cleanup, since it wasn't writing to any files, while the current version seems to be leaving the temp file behind.

Or are we trying to achieve something other than what is listed in #90?

mum4k (changes requested):

salvo/src/lib/cmd_exec.py
abaptiste: With this change I get:
```
TypeError: check_call() argument after ** must be a mapping, not CommandParameters
```

salvo/src/lib/source_tree.py
abaptiste: I can pull this out of this PR

salvo/src/lib/source_tree.py
abaptiste: I know that when the builder objects go out of scope the temporary directories get cleaned up.   I had to chase a few "File Not Found" errors due to the TemporaryDirectory getting removed.

I know I have to do some work ensuring that various artifacts are cleaned up after benchmarks are executed. We could define `__del__` in the source_tree to ensure that these objects are closed and purged, but that's a separate effort I haven't started digging into yet.

salvo/src/lib/cmd_exec.py
abaptiste: The output from some benchmarks is rather large.  This seems to work more consistently in capturing the output.

I kept this in since it did alert me to a few TemporaryDirectory objects getting prematurely reaped while addressing https://github.com/envoyproxy/envoy-perf/issues/90.

I'd prefer to keep it.

salvo/src/lib/cmd_exec.py
mum4k: Thank you trying it out.

salvo/src/lib/cmd_exec.py
mum4k: Thank you for helping me understand this. Since this qualifies as a tricky (or non-obvious) code, can we add a inline comment explaining the need for this construct? This will help future maintainers.

abaptiste: Sure thing. Let me update the PR.

salvo/src/lib/source_tree.py
mum4k: Thank you, tracking it as a separate effort is a good idea.

mum4k (approved):

PR #92, opened 2021-01-27
Add Source Manager implementation and tests

abaptiste:
This PR adds the implementation for the source manager object.  This source manager manipulates source trees and its purpose is to have a single interface through which all artifacts are compiled.

We also updated the `source.proto` to allow users to specify options that are supplied to bazel.  This allows us to build an Envoy binary that does not reject NightHawks attempts to enable profiling.

cc: @mum4k , @landesherr


salvo/src/lib/test_source_manager.py
```diff
@@ -40,23 +77,57 @@ def test_get_envoy_images_for_benchmark(mock_run_command):
...
-  Args:
-    args: the list of arguments received by the mocked function
-  """
-  pass
+  image_config = _generate_default_benchmark_images(job_control)
+  image_config.envoy_image = "envoyproxy/envoy-dev:latest"
+
+  mock_run_command.side_effect = _run_command_side_effect
+
+  src_mgr = source_manager.SourceManager(job_control)
```
mum4k, line 91: (nit) The style guide recommends against abbreviations in names, can we call this `manager` instead?

Applies to two locations.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#316-naming

salvo/src/lib/test_source_manager.py
```diff
@@ -40,23 +77,57 @@ def test_get_envoy_images_for_benchmark(mock_run_command):
   """Verify that we can determine the current and previous image
```
mum4k, line 77: (pre-existing) Should this method be called `test_get_envoy_hashes_for_benchmark` since that is the name of the method it is testing?

salvo/src/lib/test_source_manager.py
```diff
@@ -40,23 +77,57 @@ def test_get_envoy_images_for_benchmark(mock_run_command):
...
+
+  mock_run_command.side_effect = _run_command_side_effect
+
+  src_mgr = source_manager.SourceManager(job_control)
+  hashes = src_mgr.get_envoy_hashes_for_benchmark()
+
+  assert hashes == ['mocked_hash_the_sequel', 'latest']
 
 @mock.patch("src.lib.cmd_exec.run_command")
 def test_previous_hash_with_disk_files(mock_run_command):
```
mum4k, line 97: Similarly, this method should probably be called `test_get_image_hashes_from_disk_source`.

Using names that match the tested methods helps future contributors to find existing tests for a method they want to modify.

salvo/src/lib/test_source_manager.py
```diff
@@ -2,26 +2,63 @@
 Test source management operations needed for executing benchmarks
```
mum4k, line 2: We may be lacking test coverage, or have too many public methods.

Looks like the code in this PR modifies these public methods:
```python
def determine_envoy_hashes_from_source
def find_all_images_from_specified_tagsdef
def find_all_images_from_specified_sources
def get_envoy_hashes_for_benchmark
def get_image_hashes_from_disk_source
def get_source_repository
def get_source_tree
def get_build_options
def have_build_options
```

However our test file only seems to have tests for:
```python
def test_get_envoy_hashes_for_benchmark
def test_get_image_hashes_from_disk_source
```

I didn't try to generate test coverage, so might be missing something subtle. But if all the other methods need to be public, we should add test cases for them. Ideally each publicly exposed method has a test dedicated to verifying its behaviors. If some of these are just helpers that do get sufficient test coverage from the existing tests here, we can also mark them private. Or use any combination of these approaches.

On a related note, can we add a CI task that will calculate coverage on PRs? This would significantly simplify code reviews. I understand this will take some time and am wondering if you could somehow annotate PRs with the actual coverage until we get there? Alternatively I could run `blaze coverage` locally at code review time.

mum4k (changes requested):

salvo/src/lib/test_source_manager.py
abaptiste: I have a script that generates coverage locally though I haven't integrated that into any CI tasks yet.

Adding more tests is absolutely on our radar.  I can work on adding more tests and update this PR

salvo/src/lib/test_source_manager.py
abaptiste: Makes sense.  I can update.

salvo/src/lib/test_source_manager.py
abaptiste: Will update.

salvo/src/lib/test_source_manager.py
abaptiste: I'll update this

salvo/src/lib/source_manager.py
```diff
@@ -135,4 +243,115 @@ def get_source_repository(self, \
...
+    """Determine whether build options are specified in the control object
+       and return a boolean.  This is used to determine whether we build
+       images or use the already available images
+
+    Args:
+      source_id: The identity of the source object we seek (eg.
+        SRCID_NIGHTHAWK or SRCID_ENVOY)
+
+    Return:
+      the Bazel Options defined in the source identified by the
```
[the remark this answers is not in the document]
abaptiste: Yep.  Thanks catching it.

abaptiste:

salvo/src/lib/test_source_manager.py
abaptiste:
```python
$  egrep "^def test" src/lib/test_source_manager.py | cut -d\( -f 1
def test_get_envoy_hashes_for_benchmark
def test_get_image_hashes_from_disk_source
def test_determine_envoy_hashes_from_source
def test_find_all_images_from_specified_tags
def test_find_all_images_from_specified_sources
def test_get_envoy_hashes_for_benchmark
def test_get_image_hashes_from_disk_source
def test_get_source_tree
def test_get_build_options
def test_have_build_options
```

mum4k: Thank you @abaptiste for increasing the test coverage. Should we file an issue tracking addition of coverage into CI?

While we don't have it in CI, would you mind sharing the script that generates coverage? That will help me to verify coverage as part of the review process.

With that said - if filing of #93 indicates that we don't yet have full coverage for the code in this PR, let's stay here and iterate. We should prefer to submit code together with all its public behaviors tested in a single PR. That way we go from good state to good state on the head. If the scope seems too large for a single PR, we can submit less code, but still with test coverage.

abaptiste: @mum4k I added the script I have to measure coverage.  It's messy and still needs work.

If you have a better version/tool, I'm interested.

Using this script, I see the coverage for `source_manager` go from:

```
source_manager.py                         |91.3%    276|    -     0|    -      0
```

To: 
```
source_manager.py                         |97.5%    283|    -     0|    -      0
```

mum4k:
Thank you for the offline discussion about test coverage @abaptiste. Please tag me here again once it is ready for another pass. Or let me know if it already is.

abaptiste:
@mum4k I updated tests for `source_tree` and `cmd_exec`.  Both should be above 95% coverage (per the script output):
```
cmd_exec.py                                    |98.0%    51|    -   0|    -    0
source_tree.py                                 |95.6%   294|    -   0|    -    0
```
Let me know if this is good for this PR.  Thanks!

salvo/src/lib/test_source_manager.py
mum4k: Thank you very much for providing the script. I managed to adjust it a bit to generate HTML coverage reports which will help a lot while reviewing the code. We can use it until we implement coverage check in CI.

salvo/src/lib/source_manager.py
```diff
@@ -1,25 +1,47 @@
...
+
+  Args:
+    image_name: The docker image name
+
+  Return:
+    a string containing the image tag. For example:
+      envoyproxy/envoy:v1.15.3 -> v1.15.3
+  """
+  if ':' not in image_name:
+    return image_name
```
mum4k, line 41: We already have an excellent line coverage (above 99%). Normally that would be more than enough, however the two uncovered lines here seem to represent interesting behaviors. Sadly bugs usually hide in uncovered lines.

Would it be feasible to add test case to cover this line? Or remove the code if the behavior isn't useful. I will mark the other uncovered line also.

salvo/src/lib/source_manager.py
```diff
@@ -135,4 +271,114 @@ def get_source_repository(self, \
...
+    Args:
+      source_id: The identity of the source object we seek (eg.
+        SRCID_NIGHTHAWK or SRCID_ENVOY)
+
+    Return:
+      a boolean indicating the presense of user specified bazel options
+    """
+    try:
+      build_options = self.get_build_options(source_id)
+      options_present = len(build_options) >= 1
```
mum4k, line 380: This is the other uncovered line which too seems to represent non-trivial behavior. It looks like our tests only use one path through this logic - the one that raises an exception. Can we cover the non-exception path too?

mum4k (changes requested):
This was an amazing improvement in test coverage, thank you @abaptiste.

Sending what I believe is the last set of comments on the newly added tests and script.

salvo/src/lib/test_source_manager.py
```diff
@@ -35,29 +85,519 @@ def _generate_default_benchmark_images(job_control):
...
+  expected_hashes = ['v1.15.2', 'v1.16.0']
+
+  assert hashes == expected_hashes
+
+@mock.patch.object(source_tree.SourceTree, 'pull')
+@mock.patch.object(source_tree.SourceTree, 'copy_source_directory')
+def test_determine_envoy_hashes_from_source_pull_fail(mock_copy_source_directory,
+                                                      mock_source_tree_pull):
+  """
+  Verify that we can determine Envoy hashes from a source repository
```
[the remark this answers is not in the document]
abaptiste: Ack.  Will update.

salvo/src/lib/source_manager.py
abaptiste: Might be an easier path removing these lines.  I added them to avoid an `IndexError` if the split fails

salvo/tools/coverage.sh
```diff
@@ -0,0 +1,35 @@
...
+  CMD="${CMD} -a ${file}"
+done
+
+mkdir -p coverage
+CMD="${CMD} -o coverage/coverage.dat"
+(${CMD})
+
+# Extract all coverage data for salvo project files
+lcov -e coverage/coverage.dat -o coverage/salvo.dat '*/salvo/src/lib/*.py'
+
```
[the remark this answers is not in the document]
abaptiste: Nice! I hadn't considered modifying the paths in the coverage output.  I can update the script along with the other changes.  Thanks for getting this to work.

salvo/src/lib/source_manager.py
abaptiste: Updated and verified that we're hitting these lines in the html report :)

abaptiste:
@mum4k Thanks for your help.   This PR is ready for you to review.

salvo/src/lib/source_manager.py
mum4k: An exceptional test coverage ;) Thank you.

salvo/src/lib/test_source_manager.py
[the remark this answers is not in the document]
mum4k: Looks like this change wasn't performed yet. Everything else looks good.

Would you be able to update to update the description before we merge this in?

mum4k:
@abaptiste everything looks good except the remaining description change.

salvo/src/lib/test_source_manager.py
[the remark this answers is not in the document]
abaptiste: @mum4k I remember making this change. Not certain why it didn't stick.  In any case, it's updated now.  Thanks!

mum4k (approved):

PR #97, opened 2021-02-04
Update test coverage

abaptiste:
This PR syncs more test changes for Salvo.  Coverage increases from 95.6% to 97.1%

Since the docker volume. module has its own tests, we remove a redundant one from the job control loader tests.

Also, as part of the test target, we run the coverage script.

cc: @mum4k, @landesherr

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -29,7 +29,7 @@ function test_salvo() {
   pushd salvo
 
   install_deps
-  bazel test //...
+  tools/coverage.sh
 
   popd
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 **/*.pyc
+**/*.rej
+**/*.orig
 **/*.swp
 **/.vscode/*
 **/__pycache__/*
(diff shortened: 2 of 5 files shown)


ci/do_ci.sh
```diff
@@ -29,7 +29,7 @@ function test_salvo() {
   pushd salvo
 
   install_deps
-  bazel test //...
+  tools/coverage.sh
```
mum4k, line 32: Not sure what the implications for Python are, but in other languages `blaze coverage` builds in a different mode than `blaze test`, so we usually keep both modes.

This might be moot point for Python, so if you believe `blaze coverage` gives us everything we have with `blaze test`, it is probably fine.

mum4k (approved):
Looks good, just one question.

ci/do_ci.sh
abaptiste: Good question.  I was under the impression that the test modes would be congruent so running the tests under the coverage script would work the same way. as the plain "test" target.

I don't think the duplication by running "bazel test" in addition to the coverage buys us much.   I'll do some more research and hold the option to update this later.  Would that work?

mum4k: That sounds good, thank you @abaptiste.

abaptiste:
Thank you @mum4k

PR #99, opened 2021-02-09
[Salvo] Add the implementation of the envoy builder

abaptiste:
In this PR we add the implementation for the skeleton upstreamed earlier. 

We add additional changes in existing modules.  Most notably is cmd_exec, we add a second method to execute a command where we are only interested in whether the command exits successfully.

cc: @mum4k, @landesherr

### DIFF
--- a/salvo/src/lib/builder/test_envoy_builder.py
+++ b/salvo/src/lib/builder/test_envoy_builder.py
@@ -29,22 +29,20 @@ def _check_call_side_effect(args, parameters):
   if args == "bazel clean":
     return "INFO: Starting clean"
   elif args == "bazel build -c opt " + constants.ENVOY_BINARY_BUILD_TARGET:
-    return "foo"
+    return "building..."
   elif args == ("cp -fv bazel-bin/source/exe/envoy-static "
                 "build_release_stripped/envoy"):
     return "copied..."
-  elif args == (("objcopy --strip-debug bazel-bin/source/exe/envoy-static "
-                 "build_release_stripped/envoy")):
+  elif args == ("objcopy --strip-debug bazel-bin/source/exe/envoy-static "
+                "build_release_stripped/envoy"):
     return "stripped..."
-  elif args == (("docker build -f ci/Dockerfile-envoy -t "
-                 "envoyproxy/envoy-dev:v1.16.0 --build-arg "
-                 "TARGETPLATFORM='.' .")):
+  elif args == ("docker build -f ci/Dockerfile-envoy -t "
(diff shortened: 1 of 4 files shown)


salvo/src/lib/builder/test_envoy_builder.py
```diff
@@ -12,25 +12,120 @@
...
+                 "build_release_stripped/envoy")):
+    return "stripped..."
+  elif args == (("docker build -f ci/Dockerfile-envoy -t "
+                 "envoyproxy/envoy-dev:v1.16.0 --build-arg "
+                 "TARGETPLATFORM='.' .")):
+    return "docker build output..."
+
+  raise NotImplementedError(f"Unhandled arguments in call: {args}")
+
+@mock.patch.object(envoy_builder.EnvoyBuilder, 'create_docker_image')
```
mum4k, line 46: (possibly non-actionable) To be noted that the test written like this only serves as a change detector. Everything the test interacts with is a mock, so it only verifies about assumptions about the mocked objects and has no functional value.

A change detector, because anytime we change the code, we have to come here and make the very same change to the mocks. One way to address this would be to decrease the number of mocks in use, maybe at least some of the mocked objects could be real objects performing something on a fake filesystem, etc.

Feel free to just ack if that isn't feasible.

salvo/src/lib/cmd_exec.py
```diff
@@ -70,3 +70,28 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
 
   return output
 
+def run_check_command(cmd: str, parameters: CommandParameters) -> None:
+  """Run the specified command checking its exit status. If the command exits
```
mum4k, line 74: (nit) The second sentence here ("If the command exits ..."), doesn't really belong into the one line docstring since it describes a raise. The Raises section is more suited to do that.

mum4k (changes requested):

salvo/src/lib/cmd_exec.py
abaptiste: I will update this.

salvo/src/lib/builder/test_envoy_builder.py
abaptiste: I was able to remove the envoy builder mock.  It did feel grimy but I didn't circle back and take a fresh look.

My intent with the mocks was to reduce the usage and size of the side effect methods.

Side note, I did get my IDE's linter working again so you'll see other changs that it's picking up.

mum4k (approved):

PR #105, opened 2021-02-26
[Salvo] Add scavenging benchmark skeleton

abaptiste:
This PR adds the skeleton for the scavenging benchmark and its unit tests.  We also add a common module for tests containing shared object generation  methods.

cc: @mum4k, @landesherr


salvo/src/lib/benchmark/BUILD
```diff
@@ -27,14 +31,30 @@ py_test(
   ],
 )
 
+py_test(
+  name = "test_scavenging_benchmark",
```
oschaaf, line 35: So one though, I wonder if someone could come up with  a better term in favor of "scaffolding". Scavenging doesn't sound very uplifting and it is has been my bad to pick that term. No need to change it here as far as I am concerned, but maybe if we can come up with something better and feel it is worth it, we can do a global replacement.

abaptiste: Perhaps:
- Discovering Benchmark
- Probing Benchmark
- Foraging Benchmark

The last entry makes me think of bears searching for food, so that may be a something to guide the new choice for a name.

mum4k: I am assuming that if we are renaming, we are going to do it in a separate PR and proceeding with this one as is. Please let me know if that's not the case.

salvo/src/lib/benchmark/test_common.py
```diff
@@ -0,0 +1,81 @@
+"""
```
mum4k, line 1: Can we come up with a better (more specific) name for this library? If we name it `test_common`, all future test utilities will fit in here growing it to potentially large size that will be hard to support.

How about `generate_test_objects`? Feel free to choose any other more specific name.

Unrelated - isn't our testing framework going to try to execute tests in this file if we are going to keep the `test_` prefix in its name?

mum4k (changes requested):

salvo/src/lib/benchmark/BUILD
abaptiste: @mum4k You are correct.  It would be done in a separate PR.  If I understand @oschaaf correctly, the rename would be universal, not just in Salvo, but NightHawk as well.

oschaaf: If we can pick a name everyone is on happier with then I’d be up for creating PRs to replace on both sides

salvo/src/lib/benchmark/test_common.py
abaptiste: Bad naming seems to be a theme here :)  I like the 'generate_test_objects' suggestion.   This can apply elsewhere also, so I'll update.

This file is ignored by pytest because we are omitting the `if __name__ == __main__` block to run the file as a test.

salvo/src/lib/benchmark/BUILD
mum4k: @oschaaf did we choose scavenging initially because of the property where the benchmark searches for all the test files / targets to execute? Or is the word scavenging describing some other property of the benchmark?

salvo/src/lib/benchmark/test_common.py
mum4k: Thank you for the explanation about `pytest`.

mum4k (approved):

salvo/src/lib/benchmark/BUILD
oschaaf: > @oschaaf did we choose scavenging initially because of the property where the benchmark searches for all the test files / targets to execute? Or is the word scavenging describing some other property of the benchmark?

The property this was based on is that it searches/discovers test targets

mum4k: Thank you for confirming that @oschaaf.

If that is the case, it feels like neither of the proposed names explain it well, since it isn't clear **what** is being discovered or probed. How about we go with something more verbose, e.g. `discover_and_run_benchmarks` or similar.

oschaaf: `discover_and_run_benchmarks` works for me. It's a bit long but as this  would be spot on in describing what will happen, I think it won't be hard to remember.

PR #106, opened 2021-03-02
[Salvo] Add implementation for scavenging benchmark and unit tests

abaptiste:
In this PR we update the implementation and unit tests for the 'scavenging' benchmark.

There are  a few changes to the base benchmark module needed for the scavenging benchmark implementation and we tweak a few tests to increase code coverage.

cc: @mum4k, @landesherr

### DIFF
--- a/salvo/src/lib/benchmark/base_benchmark.py
+++ b/salvo/src/lib/benchmark/base_benchmark.py
@@ -1,7 +1,7 @@
 """
-Base Benchmark object module that contains
-options common to all execution methods
+Base Benchmark object module that contains common methods for all benchmarks
 """
+import abc
 import os
 import logging
 from typing import List
@@ -14,6 +14,11 @@ import api.env_pb2 as proto_env
 
 log = logging.getLogger(__name__)
 
+_VARIABLES_TO_CLEAR_AND_RESTORE = [
+ 'RUNFILES_MANIFEST_FILE'  # This variable is set by the outer bazel
+                           # invocation and negatively impacts invoking
+                           # bazel to run the scavenging benchmark
+]
 
 def get_docker_volumes(output_dir: str, test_dir: str = '') -> dict:
   """Build the volume structure needed to run a container.
@@ -34,7 +39,7 @@ class BenchmarkError(Exception):
   """Errror raised in a benchmark for an unresolvable condition."""
 
 
-class BaseBenchmark(object):
+class BaseBenchmark(abc.ABC):
   """Base Benchmark class with common functions for all invocations."""
 
   def __init__(self, job_control: proto_control.JobControl,
@@ -159,8 +164,13 @@ class BaseBenchmark(object):
     """
     return self._docker_image.run_image(image_name, run_parameters)
 
+  @abc.abstractmethod
   def execute_benchmark(self) -> None:
-    raise NotImplementedError("Method must be implemented in a derived class")
+    """Run a benchmark
+
+    Classes derived from BaseBenchmark must override execute_benchmark since the
+    execution steps differ among the supported benchmarks
+    """
 
 class BenchmarkEnvironmentError(Exception):
   """An Error raised if the environment variables required are not
@@ -169,12 +179,6 @@ class BenchmarkEnvironmentError(Exception):
 
 class BenchmarkEnvController():
   """Benchmark Environment Controller context class."""
-
-  _VARS_TO_CLEAR_AND_RESTORE = [
-     'RUNFILES_MANIFEST_FILE'  # This variable is set by the outer bazel
-                               # invocation and negatively impacts invoking
-                               # bazel to run the scavenging benchmark
-  ]
 
   def __init__(self, environment: proto_env.EnvironmentVars) -> None:
     """Initialize the environment controller with the environment object."""
@@ -214,7 +218,7 @@ class BenchmarkEnvController():
 
   def _preserve_and_clear_special_vars(self) -> None:
     """Store the name and value for any special variables."""
-    for variable in BenchmarkEnvController._VARS_TO_CLEAR_AND_RESTORE:
+    for variable in _VARIABLES_TO_CLEAR_AND_RESTORE:
       if variable in os.environ:
         self._preserved_vars[variable] = os.environ[variable]
         del os.environ[variable]
--- a/salvo/src/lib/benchmark/base_benchmark.py
+++ b/salvo/src/lib/benchmark/base_benchmark.py
@@ -7,7 +7,6 @@ import logging
 from typing import List
 
 from src.lib.docker import (docker_image, docker_volume)
-
 import api.control_pb2 as proto_control
 import api.image_pb2 as proto_image
 import api.source_pb2 as proto_source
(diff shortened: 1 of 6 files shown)


salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -105,40 +159,8 @@ def run_image(self, image_name: str,
...
-      # validation
-      if image:
-        retrieved_image = self._docker_image.pull_image(image)
-        log.debug(f"Retrieved image: {retrieved_image} for {image}")
-        if retrieved_image is None:
-          raise BenchmarkError("Unable to retrieve image: %s" % image)
-        retrieved_images.append(retrieved_image)
-
-    return retrieved_images
+  def execute_benchmark(self) -> None:
```
mum4k, line 162: (optional) A more idiomatic way of indicating that a method needs implementation on a derived class wold be to mark it as an abstract method:

https://docs.python.org/3/library/abc.html#abc.abstractmethod

mum4k, line 162: If we are expecting derived classes to implement this method, this is the location where we should have the main doc comment establishing the expected behaviors or "contract" with the child classes. This is also where we say anything specific the child classes must do in their implementations.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -148,9 +170,16 @@ class BenchmarkEnvironmentError(Exception):
 class BenchmarkEnvController():
   """Benchmark Environment Controller context class."""
 
+  _VARS_TO_CLEAR_AND_RESTORE = [
```
mum4k, line 173: Does this constant have to live on the class? Constants on classes have different lifetime which is sometimes desired, but most of the time it can confuse while debugging.

If there is nothing requiring us to package it with the class - let's just define it at the module level. We can do it just above the class to keep them close.

mum4k (changes requested):
Since we don't have coverage in CI yet, would you be able to indicate the coverage on this and any future implementation PRs?

abaptiste:
> Since we don't have coverage in CI yet, would you be able to indicate the coverage on this and any future implementation PRs?

The code coverage is run as part of the test target:
```
Overall coverage rate:
  lines......: 98.1% (3962 of 4039 lines)
  functions..: no data found
~/project

CircleCI received exit code 0
```

I can add the coverage percentage in the text of the PR.  It is on my TODO list to enforce the coverage numbers in the build process and fail on regressions.

mum4k:
Adding the coverage percentage in the text of the PR helps a lot. If you don't mind - let's do that until we have it in the CI. Thank you!

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Right, I forgot about that.  I will update.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Updated.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Updated.  Please let me know if more changes are needed, thanks!

abaptiste:
New coverage stats:

```
Writing directory view page.
Overall coverage rate:
  lines......: 98.0% (3946 of 4025 lines)
  functions..: no data found
```

abaptiste:

salvo/src/lib/benchmark/base_benchmark.py
mum4k: Since we expect the child classes to override this method, we should outline the expected "contract". I.e. what is this method supposed to do regardless of the implementation (can we say anything more specific than "it is supposed to execute the benchmark"?) and how is the method supposed to indicate the resulting status (ok or an error). E.g. is there a common exception they should be raising? Is there a common return value they should be returning (I see the return value is None, this is just an example of what "contract" means.)

Essentially a developer implementing the next child class should be able to look at this doc comment and know exactly how does the implementation have to behave in order to fulfill the API.

mum4k:
Thanks @abaptiste, it may be worth iterating one more time on the doc comment of `execute_benchmark`, otherwise this is ready to go. Left a comment in the original thread.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: I can update the comment.  We can make the method raise an exception if the benchmark fails since we are at the point where the overall logic works in Salvo.  We still see some NightHawk 'failures' where Envoy generates a `SIGTERM` at the end of a test and did not want to chase those issues while working on Salvo.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -105,40 +164,32 @@ def run_image(self, image_name: str,
...
+    A class derived from BaseBenchmark is responsible for building and staging
+    the required binaries, tests, and any other artifacts for running a
+    benchmark.
 
-    Returns:
-        a List of required image names that are retrievable. If any image is
-          unavailable, we return an empty list. The intent is for the caller to
-          build the necessary images.
+    For example in the scavenging benchmark, that class must prepare
+    the NightHawk benchamrk and binary docker image, as well as build an Envoy
```
mum4k, line 176: Typo: benchamrk.

salvo/src/lib/benchmark/base_benchmark.py
```diff
@@ -105,40 +164,32 @@ def run_image(self, image_name: str,
     """
     return self._docker_image.run_image(image_name, run_parameters)
 
-  def pull_images(self) -> List[str]:
-    """Retrieve all images necessary for the benchmark.
+  @abc.abstractmethod
+  def execute_benchmark(self) -> bool:
```
mum4k, line 168: Returning a bool lacks the ability to give a good error message. The workflow will just say "failed to execute benchmark" and the user will have to read the logs to find the actual failure.

Would it be better to change this to returning None and raising an exception instead?

salvo/src/lib/benchmark/base_benchmark.py
mum4k: Thank you, the documentation in the latest commit answered all the questions.

Resolving this comment thread, but leaving a separate comment about the return value.

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: ack.  Will fix.  I need to get spellcheck going for  Salvo :(

salvo/src/lib/benchmark/base_benchmark.py
abaptiste: Give me a few minutes :)  Let me update things.

salvo/src/lib/benchmark/base_benchmark.py
mum4k: I am haunted by the same issues as well. Thanks for updating.

mum4k (approved):

abaptiste:
Thanks again for the merge @mum4k!

PR #107, opened 2021-03-08
[Salvo] No new features cleanup and a minor bugfix.

abaptiste:
In this PR we have another round of miscellaneous fixes and items that were cleaned up.  Most of the changes revolve around string duplication.

Additional changes:
- renamed the job control build target to match the module's filename.  This keeps things a bit more congruent.
- made one bugfix where we created a set union from a string instead of a list containing that single string.
- added the source_manager as a dependency to one of the test modules.
- fixed another instance where `data` needed to be changed to `srcs`.

cc: @mum4k, @landesherr

### DIFF
--- a/salvo/src/lib/builder/test_nighthawk_builder.py
+++ b/salvo/src/lib/builder/test_nighthawk_builder.py
@@ -9,6 +9,8 @@ import api.control_pb2 as proto_control
 from src.lib.builder import nighthawk_builder
 from src.lib import (cmd_exec, constants, source_tree, source_manager)
 
+_bazel_clean_cmd = "bazel clean"
+
 @mock.patch.object(source_manager.SourceManager, 'get_source_repository')
 def test_prepare_nighthawk_source_fail(mock_get_source_tree):
   """Verify an exception is raised if the source identity is invalid"""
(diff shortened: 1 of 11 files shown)


salvo/src/lib/builder/test_nighthawk_builder.py
```diff
@@ -9,6 +9,8 @@
 from src.lib.builder import nighthawk_builder
 from src.lib import (cmd_exec, constants, source_tree, source_manager)
 
+_bazel_clean_cmd = "bazel clean"
```
mum4k, line 12: Since this is a constant, its name should be all-caps:

```python
_BAZEL_CLEAN_CMD = "bazel clean"
```

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#s3.16-naming

salvo/src/lib/docker/test_docker_image_builder.py
```diff
@@ -13,6 +13,8 @@
 import api.source_pb2 as proto_source
 import api.control_pb2 as proto_control
 
+_default_envoy_image_tag = "envoy/envoy-dev:envoy_tag"
```
mum4k, line 16: I must have missed these before, sorry about that. As constants, their names should also be all-caps.

salvo/src/lib/test_source_tree.py
```diff
@@ -9,6 +9,8 @@
 
 import api.source_pb2 as proto_source
 
+_default_https_repo_url = 'https://github.com/someawesomeproject/repo.git'
```
mum4k, line 12: Note - another constant that should be all-caps.

mum4k (changes requested):
Just one note about naming constants in multiple locations (some of them pre-existing, sorry).

salvo/src/lib/docker/test_docker_image_builder.py
mum4k: I think there are a few pre-existing ones in this file also, but we can address those in a separate PR.

Thank you for updating.

mum4k (approved):

PR #109, opened 2021-03-10
[Salvo] Add run_benchmark module to make Salvo functional

abaptiste:
With this PR, salvo becomes usable. We add the run_benchmark module which instantiates the classes needed to execute the configured benchmark.  The binary benchmark is still in-progress and will be added when it's ready.  At the moment, the Scavenging Benchmark and Fully Dockerized Benchmarks are supported.

Unit tests for the run_benchmark module are in a separate PR due to a discovered naming conflict with the `docker` python module and Salvo's `docker` module.

Coverage stats:
```
Writing directory view page.
Overall coverage rate:
  lines......: 98.0% (3963 of 4045 lines)
  functions..: no data found
```

Note: The percentage didn't drop because we don't have a test for `run_benchmark.py` in this PR.  In the test PR, the coverage data dips slightly as it stands. :( 

```
Writing directory view page.
Overall coverage rate:
  lines......: 97.8% (4369 of 4466 lines)
  functions..: no data found
```

cc: @mum4k, @landesherr

### DIFF
--- a/salvo/salvo.py
+++ b/salvo/salvo.py
@@ -3,10 +3,10 @@
 import argparse
 import logging
 import os
-import site
 import sys
 
 from src.lib.job_control_loader import load_control_doc
+from src.lib.run_benchmark import BenchmarkRunner
 
 LOGFORMAT = "%(asctime)s: %(process)d [ %(levelname)-5s] [%(module)-5s] %(message)s"
 
@@ -23,6 +23,8 @@ def setup_logging(loglevel: int=logging.DEBUG) -> None:
     loglevel configures the level of the logger.  The default is DEBUG
       level logging
   """
+  logging.getLogger('docker').setLevel(logging.ERROR)
+  logging.getLogger('urllib3').setLevel(logging.ERROR)
 
   logging.basicConfig(format=LOGFORMAT, level=loglevel)
 
@@ -38,8 +40,7 @@ def setup_options() -> argparse.Namespace:
   parser.add_argument('--job',
                       dest='jobcontrol',
                       help='specify the location for the job control json document')
-  # FIXME: Add an option to generate a default job Control JSON/YAML
-
+  # TODO: Add an option to generate a default job Control JSON/YAML
   return parser.parse_args()
 
 def main() -> int:
@@ -60,10 +61,15 @@ def main() -> int:
     return 1
 
   job_control = load_control_doc(args.jobcontrol)
+  if job_control is None:
+    log.error(f"Unable to load or parse job control: {args.jobcontrol}")
+    return 1
 
   log.debug("Job definition:\n%s\n%s\n%s\n", '=' * 20, job_control, '=' * 20)
 
   # Execute the benchmark given the contents of the job control file
+  benchmark = BenchmarkRunner(job_control)
+  benchmark.execute()
 
   return 0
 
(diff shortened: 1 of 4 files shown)


salvo/salvo.py
```diff
@@ -3,10 +3,10 @@
 import argparse
 import logging
 import os
-import site
 import sys
 
 from src.lib.job_control_loader import load_control_doc
+from src.lib.run_benchmark import BenchmarkRunner
```
mum4k, line 9: Can we change the import, so that it imports the module rather than the class?

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#22-imports

salvo/src/lib/run_benchmark.py
```diff
@@ -0,0 +1,323 @@
...
+  def _pull_or_build_nh_benchmark_image(self, \
+                                        images: proto_image.DockerImages) \
+                                        -> None:
+    """Attempt to pull the NightHawk Benchmark Image.  Build the image if
+       unavailable.
+
+       Args:
+        images: the DockerImages appearing in the control object
+    """
+    di = docker_image.DockerImage()
```
mum4k, line 119: Can we find a longer and more descriptive name instead of just `di`? Maybe `docker_image` ?

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#22-imports

salvo/src/lib/run_benchmark.py
```diff
@@ -0,0 +1,323 @@
...
+    have_build_options = self._source_manager.have_build_options(
+        proto_source.SourceRepository.SourceIdentity.SRCID_ENVOY)
+
+    envoy_images = set()
+    di = docker_image.DockerImage()
+
+    log.debug(f"Finding matching images for hashes: {image_hashes}")
+
+    for image_hash in image_hashes:
+      assert image_hash
```
mum4k, line 197: Assert statements shouldn't be used to validate arguments (I am assuming image_hashes come as an input argument or configuration from the user). Please feel free to correct me if this is purely internal validation.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#244-decision

Instead we should perform a check and raise an exception explicitly.

mum4k (changes requested):
What would we need to do to include the unit tests for this code together with this code?

If it is about renaming some modules to achieve that, we can postpone this PR and work on the rename. Reviewing the unit tests together with the code helps the reviewer to understand the intended behaviors.

salvo/salvo.py
abaptiste: Certainly.  This instance flew under the radar.

salvo/src/lib/run_benchmark.py
abaptiste: Updated

salvo/src/lib/run_benchmark.py
abaptiste: This can be removed.  Likely a leftover artifact from debugging.  I haven't seen this assert fail.

abaptiste:
> What would we need to do to include the unit tests for this code together with this code?
> 
> If it is about renaming some modules to achieve that, we can postpone this PR and work on the rename. Reviewing the unit tests together with the code helps the reviewer to understand the intended behaviors.

OK, let me close this PR, submit one for the renaming, and then we can have one PR for the run_benchmark and its tests.

mum4k:
Thank you for accommodating the request. Feel free to reopen this one if it suits, since it already has part of our review history on the code in question.

PR #111, opened 2021-03-11
[Salvo] Add Run Benchmark and Unit Tests

abaptiste:
This PR adds the run_benchmark module and its unit tests.   We incorporate a few changes to address a separate PR's comments.

In this PR we move the test object generate module into `src/lib` so that other tests can use it.

Coverage (line count is up, percentage stays the same):
```
Writing directory view page.
Overall coverage rate:
  lines......: 98.0% (4308 of 4398 lines)
  functions..: no data found
```

cc:  @mum4k, @landesherr


salvo/src/lib/run_benchmark.py
```diff
@@ -0,0 +1,324 @@
...
+        self._test.append(benchmark)
+
+    elif self._control.binary_benchmark:
+      current_benchmark_name = "Binary Benchmark"
+      job_control_list = []
+
+    if not self._test:
+      raise NotImplementedError(f"No [{current_benchmark_name}] defined")
+
+  def generate_job_control_for_envoy_images(self) \
```
mum4k, line 87: This class has a few more public methods (other than execute). Assuming they do need to be public, can we also add unit tests for them, so that every publicly available method has its own set of tests that verify it?

mum4k (changes requested):
Looks good, just one note about tests.

salvo/src/lib/run_benchmark.py
abaptiste: I can make them private.  They aren't really useful when called externally to the module

mum4k (approved):

PR #113, opened 2021-03-25
Add Binary Benchmark Workflow to Salvo

landesherr:
This PR adds the 3rd and final benchmark style, the BinaryBenchmark,
which builds Envoy and Nighthawk from source, and tests the native
binaries, rather than docker images.

Signed-off-by: Nolan Varani <landesherr@users.noreply.github.com>

cc: @mum4k @abaptiste

### DIFF
--- a/salvo/src/lib/benchmark/binary_benchmark.py
+++ b/salvo/src/lib/benchmark/binary_benchmark.py
@@ -100,7 +100,7 @@ class Benchmark(base_benchmark.BaseBenchmark):
       # We don't have the necessary sources or binaries to run Envoy,
       # but we can proceed using Nighthawk's test server as a fallback.
       log.warning("Skipping Envoy preparation - no sources/binaries were provided."
-        "Will proceed with Nighthawk test server.")
+                  "Will proceed with Nighthawk test server.")
       return
     if self.envoy_binary_path:
       if not (os.path.exists(self.envoy_binary_path) \
@@ -108,7 +108,7 @@ class Benchmark(base_benchmark.BaseBenchmark):
         # If an "Envoy" is specified, but is not a valid executable file,
         # try to proceed anyway using Nighthawk's test server.
         log.warning("ENVOY_PATH environment variable specified, but invalid."
-          "Falling back to Nighthawk test server.")
+                    "Falling back to Nighthawk test server.")
         self.use_fallback = True
         return
       # We already have a binary, no need to build
@@ -132,18 +132,18 @@ class Benchmark(base_benchmark.BaseBenchmark):
 
     #todo: refactor args, have frontend specify them via protobuf
     cmd = ("bazel test "
-          "--test_summary=detailed "
-          "--test_output=all "
-          "--test_arg=--log-cli-level=info "
-          "--test_env=ENVOY_IP_TEST_VERSIONS=v4only "
-          "--test_env=HEAPPROFILE= "
-          "--test_env=HEAPCHECK= "
-          "--cache_test_results=no "
-          "--compilation_mode=opt "
-          "--cxxopt=-g "
-          "--cxxopt=-ggdb3 "
-          "--define tcmalloc=gperftools "
-          "//benchmarks:* ")
+           "--test_summary=detailed "
+           "--test_output=all "
+           "--test_arg=--log-cli-level=info "
+           "--test_env=ENVOY_IP_TEST_VERSIONS=v4only "
+           "--test_env=HEAPPROFILE= "
+           "--test_env=HEAPCHECK= "
+           "--cache_test_results=no "
+           "--compilation_mode=opt "
+           "--cxxopt=-g "
+           "--cxxopt=-ggdb3 "
+           "--define tcmalloc=gperftools "
+           "//benchmarks:* ")
 
     cmd_params = cmd_exec.CommandParameters(cwd=self._benchmark_dir)
 
--- a/salvo/src/lib/benchmark/binary_benchmark.py
+++ b/salvo/src/lib/benchmark/binary_benchmark.py
@@ -1,0 +1,174 @@
+"""
+This module contains the methods to perform a Binary benchmark using
+containers for the scripts, nighthawk binaries, and envoy
+
+https://github.com/envoyproxy/nighthawk/blob/master/benchmarks/README.md
+"""
+
+import subprocess
+import logging
+import os
+
+import api.control_pb2 as proto_control
+import api.source_pb2 as proto_source
+
+from src.lib.benchmark import base_benchmark
+from src.lib.builder import (envoy_builder, nighthawk_builder)
+from src.lib import (cmd_exec, source_manager)
+
+log = logging.getLogger(__name__)
+
+
+class BinaryBenchmarkError(Exception):
+  """Error raised when running a binary benchmark in cases
+     where we cannot make progress due to abnormal conditions.
+  """
+
+class Benchmark(base_benchmark.BaseBenchmark):
+  """This benchmark class is the binary benchmark. We use a path to an Envoy
+     binary to execute the Nighthawk benchmarks using that specific build.
+  """
+
+  def __init__(
+    self, job_control: proto_control.JobControl, benchmark_name: str) -> None:
+    """ Initialize the benchmark class."""
+
+    super(Benchmark, self).__init__(job_control, benchmark_name)
+    self._benchmark_dir = None
+    self.envoy_binary_path = job_control.environment.variables['ENVOY_PATH']
+    self.envoy_builder = None
+    self.nighthawk_builder = None
+    self.source_manager = source_manager.SourceManager(job_control)
+    self.use_fallback = False
+
+  def _validate(self) -> None:
+    """Validate that all data required for running a benchmark exists.
+
+    Verify that a source has been specified with which to build Nighthawk,
+    warns the user if no valid Envoy binary is specified.
+
+    Returns:
+        None
+    """
+
+    source = self.get_source()
+    if not source:
+      raise BinaryBenchmarkError("No source configuration specified")
+
+    can_build_envoy = False
+    can_build_nighthawk = False
+
+    for source_def in source:
+      if source_def.identity == source_def.SRCID_UNSPECIFIED:
+        raise BinaryBenchmarkError("No source identity specified")
+
+      if source_def.identity == source_def.SRCID_ENVOY and \
+        (source_def.source_path or source_def.source_url):
+        can_build_envoy = True
+
+      if source_def.identity == source_def.SRCID_NIGHTHAWK and \
+        (source_def.source_path or source_def.source_url):
+        can_build_nighthawk = True
+
+    if not can_build_nighthawk:
+      raise BinaryBenchmarkError("No source specified to build Nighthawk")
+
+    if not (can_build_envoy or self.envoy_binary_path):
+      self.use_fallback = True
+      log.warning("No Envoy source or binary was specified. Falling back to Nighthawk test server.")
+
+
+  def _prepare_nighthawk(self) -> None:
+    """Prepare the nighthawk source for the benchmark.
+
+    Checks out the nighthawk source if necessary, builds the client
+    and server binaries
+
+    """
+
+    self.nighthawk_builder = nighthawk_builder.NightHawkBuilder(self.source_manager)
+    self.nighthawk_builder.build_nighthawk_binaries()
+    self.nighthawk_builder.build_nighthawk_benchmarks()
+
+    nighthawk_source = self.source_manager.get_source_tree(
+        proto_source.SourceRepository.SourceIdentity.SRCID_NIGHTHAWK
+    )
+    self._benchmark_dir = nighthawk_source.get_source_directory()
+
+  def _prepare_envoy(self) -> None:
+    if self.use_fallback:
+      # We don't have the necessary sources or binaries to run Envoy,
+      # but we can proceed using Nighthawk's test server as a fallback.
+      log.warning("Skipping Envoy preparation - no sources/binaries were provided."
+        "Will proceed with Nighthawk test server.")
+      return
+    if self.envoy_binary_path:
+      if not (os.path.exists(self.envoy_binary_path) \
+          and os.access(self.envoy_binary_path, os.X_OK)):
+        # If an "Envoy" is specified, but is not a valid executable file,
+        # try to proceed anyway using Nighthawk's test server.
+        log.warning("ENVOY_PATH environment variable specified, but invalid."
+          "Falling back to Nighthawk test server.")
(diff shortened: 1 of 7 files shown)


mum4k:
@abaptiste considering the amount of experience you have developing Salvo, would you be able to do the first round of review on this PR? I'll be happy to review after you indicate this looks good to you.

Please let me know in case this doesn't work, but if you do have the cycles to review - your help is appreciated.

salvo/src/lib/benchmark/binary_benchmark.py
```diff
@@ -0,0 +1,174 @@
...
+        proto_source.SourceRepository.SourceIdentity.SRCID_NIGHTHAWK
+    )
+    self._benchmark_dir = nighthawk_source.get_source_directory()
+
+  def _prepare_envoy(self) -> None:
+    if self.use_fallback:
+      # We don't have the necessary sources or binaries to run Envoy,
+      # but we can proceed using Nighthawk's test server as a fallback.
+      log.warning("Skipping Envoy preparation - no sources/binaries were provided."
+        "Will proceed with Nighthawk test server.")
```
abaptiste, line 103: nit:  indentation update needed here

salvo/src/lib/benchmark/binary_benchmark.py
```diff
@@ -0,0 +1,174 @@
...
+      return
+    if self.envoy_binary_path:
+      if not (os.path.exists(self.envoy_binary_path) \
+          and os.access(self.envoy_binary_path, os.X_OK)):
+        # If an "Envoy" is specified, but is not a valid executable file,
+        # try to proceed anyway using Nighthawk's test server.
+        log.warning("ENVOY_PATH environment variable specified, but invalid."
+          "Falling back to Nighthawk test server.")
```
abaptiste, line 111: nit: same indentation update needed here

salvo/src/lib/benchmark/binary_benchmark.py
```diff
@@ -0,0 +1,174 @@
...
+    specified source.
+    """
+
+    self._validate()
+    self._prepare_nighthawk()
+    self._prepare_envoy()
+
+    #todo: refactor args, have frontend specify them via protobuf
+    cmd = ("bazel test "
+          "--test_summary=detailed "
```
abaptiste, line 135: nit: Single space needed here and the following lines.

salvo/src/lib/benchmark/test_binary_benchmark.py
```diff
@@ -0,0 +1,183 @@
...
+
+  # Calling execute_benchmark should throw an exception
+  with pytest.raises(binary_benchmark.BinaryBenchmarkError) \
+      as validation_error:
+    benchmark.execute_benchmark()
+
+  assert str(validation_error.value) == "No source configuration specified"
+
+# Mock the build invocations so we don't actually try to build this big chungus
+@patch('src.lib.builder.nighthawk_builder.NightHawkBuilder.build_nighthawk_benchmarks')
```
abaptiste, line 35: Can we make these a string constant above? The lines herein all tests  exceed 80 characters.

salvo/src/lib/run_benchmark.py
```diff
@@ -7,9 +7,8 @@
 import os
 from typing import (List, Set)
 
-from src.lib.benchmark import fully_dockerized_benchmark as fulldocker
-from src.lib.benchmark import scavenging_benchmark as scavenging
-from src.lib.benchmark import base_benchmark
+from src.lib.benchmark import (fully_dockerized_benchmark as fulldocker,
+  scavenging_benchmark as scavenging, binary_benchmark as binbench, base_benchmark)
```
abaptiste, line 11: nit: pylint indentation again

salvo/src/lib/run_benchmark.py
```diff
@@ -116,11 +194,10 @@ def _pull_or_build_nh_benchmark_image(self, \
        Args:
         images: the DockerImages appearing in the control object
     """
-
+    di = docker_image.DockerImage()
```
abaptiste, line 197: We've changed `di` in previous PRs to use a more descriptive name.  The renamed variable is `image_manager` IIRC.  Can we do that here?

salvo/src/lib/run_benchmark.py
```diff
@@ -191,18 +267,19 @@ def _pull_or_build_envoy_images_for_benchmark(
         proto_source.SourceRepository.SourceIdentity.SRCID_ENVOY)
 
     envoy_images = set()
-    image_manager = docker_image.DockerImage()
+    di = docker_image.DockerImage()
 
     log.debug(f"Finding matching images for hashes: {image_hashes}")
 
     for image_hash in image_hashes:
+      assert image_hash
```
abaptiste, line 275: This assert should be removed.

salvo/src/lib/test_run_benchmark.py
```diff
@@ -151,7 +167,7 @@ def test_benchmark_failure_if_no_benchmark_selected():
     _ = run_benchmark.BenchmarkRunner(job_control)
 
   assert str(not_implemented.value) == \
-      "No [Unspecified Benchmark] defined"
+      "No [Unspecified Benchmark] defined yet"
```
abaptiste, line 170: Can you double check that we aren't overwriting changes from previous PR's?  This is another instance where we are reverting changes already upstreamed.

abaptiste (changes requested):
Seems that the diff is reverting some upstreamed changes.  Can you make another pass and double check the `git diff` output?  Thanks!

salvo/src/lib/benchmark/binary_benchmark.py
landesherr: Use 4-space indentation for multi-line statements? Or align with the start of the quote on the previous line?

salvo/src/lib/test_run_benchmark.py
landesherr: There were some merge conflicts between our copy of the repo and what has been upstreamed. For some reason, I thought it was new stuff that we had wanted to upstream, but I was wrong about that.

Should be all fixed now though :)

abaptiste (approved):
LGTM.  @mum4k, this is ready for a pass from you.  Thanks!

salvo/src/lib/run_benchmark.py
```diff
@@ -107,6 +111,53 @@ def _generate_job_control_for_envoy_images(self) \
...
 
+  def _generate_job_control_for_binaries(self) \
+    -> List[proto_control.JobControl]:
+    """Determine the required source configurations needed for a binary benchmark.
+
+    Find the specified version of Nighthawk, Envoy, and all additional Envoy
+    commits to be tested. These will be spawned into multiple Binary Benchmark jobs.
+
+    Raises:
+      Exception: when multiple competing repositories of Envoy/Nighthawk are specified.
```
mum4k, line 122: (partly pre-existing) Can we raise our custom exception here (`BenchmarkRunnerError`) instead of the top level `Exception`?

Raising the top level `Exception` causes the users to catch at that level which would also catch and cover any other Python exception, e.g. SyntaxError or MemoryError that might not be desired.

mum4k (changes requested):
Thank you @abaptiste and @landesherr for the initial review rounds. Left a few comments and questions.

salvo/src/lib/benchmark/binary_benchmark.py
```diff
@@ -0,0 +1,174 @@
...
+     binary to execute the Nighthawk benchmarks using that specific build.
+  """
+
+  def __init__(
+    self, job_control: proto_control.JobControl, benchmark_name: str) -> None:
+    """ Initialize the benchmark class."""
+
+    super(Benchmark, self).__init__(job_control, benchmark_name)
+    self._benchmark_dir = None
+    self.envoy_binary_path = job_control.environment.variables['ENVOY_PATH']
```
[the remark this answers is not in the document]
landesherr: These do not need to be public. I will go ahead and make them private.

salvo/src/lib/benchmark/binary_benchmark.py
```diff
@@ -0,0 +1,174 @@
...
+
+      if source_def.identity == source_def.SRCID_NIGHTHAWK and \
+        (source_def.source_path or source_def.source_url):
+        can_build_nighthawk = True
+
+    if not can_build_nighthawk:
+      raise BinaryBenchmarkError("No source specified to build Nighthawk")
+
+    if not (can_build_envoy or self.envoy_binary_path):
+      self.use_fallback = True
```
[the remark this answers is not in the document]
landesherr: Now that I'm taking another look at the fallback functionality, I'm starting to realize that removing it altogether is the best course of action. Technically, nighthawk *can* execute without an envoy binary, using its built-in test server, but that doesn't give any tangible benefit for the kinds of workloads that Salvo is expected to automate.

Additionally, the frontend (i.e. `run_benchmark.py`) will already fail-fast when no Envoy source is provided.

This is essentially a vestigial feature from the initial development of the Binary Benchmark workflow, and your comments highlight to me that it is more appropriate to fail the workflow in this case.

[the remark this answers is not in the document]
oschaaf: Re: Now that I'm taking another look at the fallback functionality, I'm starting to realize that removing it altogether is the best course of action. Technically, nighthawk can execute without an envoy binary, using its built-in test server, but that doesn't give any tangible benefit for the kinds of workloads that Salvo is expected to automate.

Drive by comment: it might serve a purpose in baselining?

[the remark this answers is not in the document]
landesherr: > it might serve a purpose in baselining?

That is a possibility, though any particular version of Envoy could also be used as a baseline.
If we want to keep the fallback logic, I'd probably have it be controlled by an environment variable in the job control.

If this sounds like a useful feature, I can add it - but I kind of doubt that people would use it at the moment.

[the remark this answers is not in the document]
oschaaf: So, to illustrate, with baseline I meant `nh client -> nh test server` and then one can compare against `nh client -> envoy -> nh test server`. To see how introducing an Envoy proxy affects measurements.

If it's extra work, that's not my intent, it's just that the comment stood out, and it seemed like something would be axed out that I thought could serve a purpose.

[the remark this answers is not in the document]
landesherr: I see. That does make sense, and I can see now how that would add value.
Unfortunately, this feature as it exists is not complete on an end-to-end basis, and would require additional development and testing on the frontend/job control side in order to get it working.

I think this is a good callout, but I think the best course of action is to remove it for now.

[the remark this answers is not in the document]
oschaaf: I would rely on you for judging that, but perhaps it is good to file an issue for it so it can be tracked.

The benchmark scripts support the flow, for example:
https://github.com/envoyproxy/nighthawk/blob/main/benchmarks/test/test_discovery.py#L101-L105
(which produces https://27055-180498819-gh.circle-artifacts.com/0/2021-03-29-18-55-03/http_h1_small_request_small_reply_direct_IpVersion.IPV4-nighthawk_test_integration_configurations_nighthawk_http_origin.yaml/simple_plot.html in CI)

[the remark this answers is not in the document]
oschaaf: Sorry, thinking about this more, one afterthought.
The above discussion represents a case where there's 0 Envoy instances.
If it impacts things here, >1 Envoy instances (from the same version/binary) might be good to consider too in the context of this: a chain of Envoy's is a realistic scenario, for example gateway->sidecar, and therefore could be interesting to test at some point.

salvo/src/lib/benchmark/binary_benchmark.py
```diff
@@ -0,0 +1,178 @@
...
+  def __init__(
+    self, job_control: proto_control.JobControl, benchmark_name: str) -> None:
+    """ Initializes the benchmark class
+
+    Args:
+        job_control: The protobuf object containing the parameters and locations
+          of benchmark artifacts
+        benchmark_name: The name of the benchmark to execute
+
+    Raises:
```
mum4k, line 41: This Raises section may not be correct, since the initializer doesn't contain any `raise` statements. We only document raises that are directly invoked by the function we are describing.

salvo/src/lib/benchmark/binary_benchmark.py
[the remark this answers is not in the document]
mum4k: Agreed that dropping this functionality from this PR is a good idea. @oschaaf thank you for raising a good point about the need for having a baseline check for the Salvo/Nighthawk combo itself. Otto, if you have the cycles, I would ask for your help with filing of the issue describing this functionality (both 0 Envoys and >1 Envoys). While the functionality isn't on the critical path that we are currently developing, it is something worth exploring and tracking in the future.

salvo/src/lib/run_benchmark.py
```diff
@@ -9,7 +9,9 @@
 
 from src.lib.benchmark import fully_dockerized_benchmark as fulldocker
 from src.lib.benchmark import scavenging_benchmark as scavenging
+from src.lib.benchmark import binary_benchmark as binbench
```
mum4k, line 12: Sorry I should have noted this in the previous review round. We also shouldn't rename imports, the style guide only calls out two reasons for renaming - when they are inconveniently long and when there are two imports with the same name. These don't seem to meet either of the exceptions. Keeping import names helps the reader to quickly find the module without the indirection.

https://github.com/google/styleguide/blob/gh-pages/pyguide.md#224-decision

mum4k (changes requested):

salvo/src/lib/run_benchmark.py
landesherr: Would you like me to change all of the benchmark import names while we're here? Or is that out of scope for this PR?

mum4k: It might be best not to extend the scope of this PR. We can either just change the one we are adding, or we can file an issue to change all of them later.

landesherr: > It might be best not to extend the scope of this PR. 

Agreed. It's nice to have fairly atomic commits in the history, IMO.
Since I'm already going to follow up with that doc fix, I'll change it for the binary benchmark, and I can create an issue for the rest of the reformatting in `run_benchmark.py`

landesherr: Issue created: https://github.com/envoyproxy/envoy-perf/issues/114

mum4k (approved):

abaptiste:
Thanks for the merge @mum4k!

mum4k:
My pleasure, thanks goes to the both of you for contributing the binary benchmark!

Issue #122, opened 2021-11-10
Can you please give a more detailed readme of how to use salvo?


jiangdongzi:
I do not know how what files should put in the `testDir`  of `Fully Dockerized Benchmark`. Hope you can give an example.

mum4k:
Hi @jiangdongzi, you raise a good point about the lack of detailed documentation. The development of the Salvo project has been paused temporarily. We expect to get to it in the future, at which point we will also provide better documentation on how to use it.

I am happy to provide answers meanwhile. The `testDir` is optional and can contain tests defined by the user. Salvo runs the [Nighthawk benchmarks](https://github.com/envoyproxy/nighthawk/tree/main/benchmarks) implementation. So if the user wanted to provide their own test cases, they would put test files like [this one](https://github.com/envoyproxy/nighthawk/blob/main/benchmarks/test/test_discovery.py) into the `testDir`.

Hope this helps.

gyohuangxin:
Hi @mum4k , I'm looking at this project these days and I think it'll be very helpful for me and envoy project. Do you have any updates about Salvo itself and any plan to integrate it into Envoy CI system? Thanks.

mum4k:
Thank you for reaching out @gyohuangxin, we have a fairly concrete plan at this point. We are aiming to implement a solution meeting [these requirements](https://docs.google.com/document/d/1mAma-ksRN0OIBInoZKUdjdaIhK2nTQxFnCujq2fKi4E/edit) which includes integration into Envoy's CI. We expect to have a person full time on this within the next 6 months.

Happy to discuss any of the details and/or accept any help you may be willing to offer.

gyohuangxin:
@mum4k Thank you for the answer.

gyohuangxin:
@mum4k I'm going to draft a document about the local execution mode of Salvo, can you assign this issue to me? Thanks.

mum4k:
Thank you for your help @gyohuangxin.

gyohuangxin:
@mum4k I raise a PR to revise the doc of the local execution mode of Salvo, it's not perfect, any comments from you would of great help. Thanks!

gyohuangxin:
cc @jiangdongzi I'll also appreciate it if you have any comments.

gyohuangxin:
@mum4k I raised a PR to add another detailed documentation 'Measure Envoy's Performance Change with an A/B Testing', can you review and leave some comments?

mum4k:
[...]

PR #115, opened 2021-04-08
[Salvo] Binary benchmark updates

abaptiste:
Update the README.md to contain an example of the Job Control
for a binary benchmark.  Also update the binary benchmark to
operate on the specified commit and one prior as is inline with
other implemented benchmarks

Signed-off-by: abaptiste <abaptiste@users.noreply.github.com>

cc: @mum4k

PR #126, opened 2022-01-25
[Salvo] Document of the local execution mode of Salvo (#122)

gyohuangxin:

### DIFF
--- a/salvo/README.md
+++ b/salvo/README.md
@@ -1,12 +1,35 @@
 # Salvo
 
-This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using [NightHawk](https://github.com/envoyproxy/nighthawk).
+## What is Salvo
 
-## Example Control Documents
+This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using [Nighthawk](https://github.com/envoyproxy/nighthawk).
 
-The control document defines the data needed to execute a benchmark. At the moment, we support the fully dockerized benchmark and the scavenging benchmark.  The work for the binary benchmark is in progress.
+## Goals of Salvo
+
+Salvo allows Envoy developers to perform A/B testing to monitor performance change of Envoy. Salvo provides the local excution mode allowing developers to run a benchark on their own machine and also provides the remote excution mode to run a benchmark on a remote machine, such as a remote CI system.
+
+## Dependencies
+
+The [`install_deps.sh`](./install_deps.sh) script can be used to install any dependencies required by Salvo.
+
+## Building Salvo
+
+To build Salvo, use the following command:
+
+```bash
+bazel build //...
+```
+
+## Benchmark Test Cases for Salvo
+
+Benchmark test cases for Salvo are defined as Python files with test cases written in pytest framework, here is an example: https://github.com/envoyproxy/nighthawk/blob/main/benchmarks/test/test_discovery.py. Users can provide thier own Python files of test cases into Salvo.
+
+## Control Documents
+
+The control document defines the data needed to execute a benchmark. We support the fully dockerized benchmark, the scavenging benchmark and the binary benchmark. 
 
 ### Fully Dockerized Benchmark
+
 The Fully Dockerized Benchmark discoveres user supplied tests for execution and uses docker images to run the tests. In the example below, the user supplied tests files are located in `/home/ubuntu/nighthawk_tests` and are mapped to a volume in the docker container.
 
 To run the dockerized benchmark, create a file with the following example contents:
@@ -21,7 +44,7 @@ JSON Example:
     "reuseNhImages": true,
     "nighthawkBenchmarkImage": "envoyproxy/nighthawk-benchmark-dev:latest",
     "nighthawkBinaryImage": "envoyproxy/nighthawk-dev:latest",
-    "envoyImage": "envoyproxy/envoy:v1.16.0"
+    "envoyImage": "envoyproxy/envoy:v1.21.0"
   },
   "environment": {
     "testVersion": IPV_V4ONLY,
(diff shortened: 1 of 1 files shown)


salvo/README.md
```diff
@@ -1,12 +1,35 @@
...
 This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using [NightHawk](https://github.com/envoyproxy/nighthawk).
 
-## Example Control Documents
+## Goals of Salvo
+
+Salvo allows Envoy developers to perform A/B testing to monitor performance change of Envoy. Salvo provides the local excution mode allowing developers to run a benchark on their own machine and also provides the remote excution mode to run a benchmark on a remote machine, such as a remote CI system.
+
+## Dependencies
+
+The `install_deps.sh` script can be used to install any dependencies required by Salvo.
```
mum4k, line 13: Can this be a link to the script?

salvo/README.md
```diff
@@ -1,12 +1,35 @@
...
+
+To build Salvo, use the following command:
+
+```bash
+bazel build //...
+```
+
+## Benchmark Test Cases for Salvo
+
+Benchmark test cases for Salvo are defined as Python files with test cases written in pytest framework, here is an example: https://github.com/envoyproxy/nighthawk/blob/main/benchmarks/test/test_discovery.py. Users can provide thier own Python files of test cases into Savlo.
```
mum4k, line 25: Typo: `Savlo`.

salvo/README.md
```diff
@@ -38,21 +61,42 @@ YAML Example:
...
 images:
   reuseNhImages: true
   nighthawkBenchmarkImage: 'envoyproxy/nighthawk-benchmark-dev:latest'
   nighthawkBinaryImage: 'envoyproxy/nighthawk-dev:latest'
-  envoyImage: "envoyproxy/envoy:v1.16.0"
+  envoyImage: "envoyproxy/envoy:v1.21.0"
 ```
 
-In both examples above, the envoy image being tested is a specific tag. This tag can be replaced with "latest" to test the most recently created image against the previous image built from the prior tag. If a commit hash is used, we find the previous commit hash and benchmark that container.  In summary, tags are compared to tags, hashes are compared to hashes.
+`remote`: Whether to enable remote excution mode.
```
mum4k, line 75: Can we provide few more details about what is meant by remote execution? What changes when we configure this as remote? What remote environments are supported?

salvo/README.md
```diff
@@ -1,12 +1,35 @@
...
+```
+
+## Benchmark Test Cases for Salvo
+
+Benchmark test cases for Salvo are defined as Python files with test cases written in pytest framework, here is an example: https://github.com/envoyproxy/nighthawk/blob/main/benchmarks/test/test_discovery.py. Users can provide thier own Python files of test cases into Savlo.
+
+## Control Documents
 
-The control document defines the data needed to execute a benchmark. At the moment, we support the fully dockerized benchmark and the scavenging benchmark.  The work for the binary benchmark is in progress.
+The control document defines the data needed to execute a benchmark. At the moment, we support the fully dockerized benchmark and the scavenging benchmark. The work for the binary benchmark is in progress.
```
mum4k, line 29: I was under the impression that the binary benchmark was also implemented. See:

- https://github.com/envoyproxy/envoy-perf/pull/113
- https://github.com/envoyproxy/envoy-perf/pull/115

This seems important, since the binary benchmark is likely going to be the one most useful for Envoy developers.

salvo/README.md
```diff
@@ -78,10 +122,19 @@ source:
   source_url: https://github.com/envoyproxy/nighthawk.git
 ```
 
-In this example, the v1.16.0 Envoy tag is pulled and an Envoy image generated where the Envoy binary has profiling enabled.  The user may specify option strings supported by bazel to adjust the compilation process.
+`scavengingBenchmark`: It will run scavenging benchmarks.
+
+`source.identity`: Specify whether this source location is Envoy or NightHawk.
+
+`source.commit_hash`: Specify a commit hash if applicable. If not specified we will determine this from the source tree. We will also use this field to identify the corresponding NightHawk or Envoy image used for the benchmark.
```
mum4k, line 129: (nit) We seem to have a couple of ways how we spell Nighthawk in this document. Would it be worthwhile to unify them and use just one format?

mum4k (changes requested):
This looks great and is definitely a step in the right direction when it comes to improving documentation.

I am assuming that this is just the first step towards completing this deliverable from our list of goals:

```
Document and finalize the local execution mode of Salvo,
allowing users to run a benchmark on their local machine.
```

The idea behind this deliverable is to provide a document specifically aimed at the Envoy developers. A document that will tell a story of a developer who has a pending PR in the Envoy codebase and wants to measure the performance at that commit and then compare it to the head.

This document doesn't exactly hit that goal, since it is fairly generic. An Envoy developer trying to use Salvo based on this document would still need to spend quite some time filling up the blanks. My hopes are that we can come up with a document that will be self-contained, focused on this use case and contain good introduction of the scenario and the exact commands to run together with an explanation of how to interpret the results.

Happy to go forward with this PR, just making sure we are on the same page with regards to the goal. Please feel free to reach out to me on Slack if you would like to discuss this in more detail.

gyohuangxin:
> This document doesn't exactly hit that goal, since it is fairly generic. An Envoy developer trying to use Salvo based on this document would still need to spend quite some time filling up the blanks. My hopes are that we can come up with a document that will be self-contained, focused on this use case and contain good introduction of the scenario and the exact commands to run together with an explanation of how to interpret the results.

@mum4k Yes, this README.md file is generic for new starters.  This PR aims to add more details to it. 
Thank you for clarification of our goals, I can raise another PR for another detailed doc describing the workflow of envoy development. What do you think?

salvo/README.md
gyohuangxin: It's a good idea, changed it to a link.

salvo/README.md
gyohuangxin: Thank you for the information, fixed.

salvo/README.md
gyohuangxin: Thanks. I've unified them to `Nighthawk`.

salvo/README.md
gyohuangxin: Remote mode has not been implemented yet, refer to:
https://github.com/envoyproxy/envoy-perf/blob/ed970c5986f662484920a8d7b1e40c7c29a48363/salvo/src/lib/run_benchmark.py#L397-L415

Do you know any more about the original target of `remote mode`?

mum4k:
Thank you @gyohuangxin, raising another PR with an Envoy specific doc sounds great!

salvo/README.md
mum4k: I don't know what the original intent was, but maybe we can ask @abaptiste? I suspect it was related to running on VMs in CI or similar.

Either way, the remote execution mode is irrelevant for our immediate goal (documenting the local execution). We can revisit the remote mode once we will be working on running Salvo in CI.

mum4k (approved):
Thank you, looks great!

salvo/README.md
gyohuangxin: @mum4k Thanks, got it. I‘ll focus on local execution at this stage.

salvo/README.md
gyohuangxin: > This seems important, since the binary benchmark is likely going to be the one most useful for Envoy developers.

@mum4k I'm drafting the doc about the workflow of Envoy deveploers. In my opinion, as a Envoy developer, the binary benchmark is the fastest way to test his/her own codes changes whithout building a image. Is this reason the same as yours? Can you leave your thoughts on it and about why not use the Dockerized/Scavenging Benchmark? Thanks

salvo/README.md
abaptiste: Right, remote mode has not been implemented.  We intended to build a mode where Salvo uploads all required artifacts to a remote service where the benchmarks would be executed in parallel in isolated environments.

PR #128, opened 2022-02-09
[Salvo] Add further information and remove unnecessary fields about Scavenging Benchmark in README.md

gyohuangxin:
I'm diving into using Salvo's Scavenging Benchmark, and I found some unnecessary fields that need to be removed. 


mum4k (approved):
Thank you for keeping the documentation accurate.

PR #133, opened 2022-04-09
[Salvo] Add single test mode in Salvo

### LINKED ISSUE #132: [Salvo] Add single test mode in Salvo
Single test mode is not supported in Salvo, therefore, you must run two images/commits at least when using Salvo. It's not helpful to the  scenario that someone want to test different configuration based on one specific envoy image/commit.

Take https://github.com/envoyproxy/envoy/pull/20148 as an example, it add an option `envoy.reloadable_features.http1_lazy_read_disable` in envoy configuration , and we want to use Salvo to test the performance difference when this option is ture or false based on one specific envoy image/commit. It's not necessary to find and test its previous image/commit.

We can add an option `single_test_mode` in Salvo job control configuration.

gyohuangxin:
Fix #132. Add `test_single_image` and `test_single_commit` option to enable single test mode, which is helpful to the scenario that someone want to test different configuration based on one specific envoy image/commit.

### DIFF
--- a/salvo/README.md
+++ b/salvo/README.md
@@ -2,19 +2,26 @@
 
 ## What is Salvo
 
-This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using [Nighthawk](https://github.com/envoyproxy/nighthawk).
+This is a framework that abstracts executing multiple benchmarks of the Envoy Proxy using 
+[Nighthawk](https://github.com/envoyproxy/nighthawk).
 
 ## Goals of Salvo
 
-Salvo allows Envoy developers to perform A/B testing to monitor performance change of Envoy. Salvo provides the local execution mode allowing developers to run a benchmark on their own machine and also provides the remote execution mode to run a benchmark on a remote machine, such as a remote CI system.
+Salvo allows Envoy developers to perform A/B testing to monitor performance change of Envoy. Salvo
+provides the local execution mode allowing developers to run a benchmark on their own machine and
+also provides the remote execution mode to run a benchmark on a remote machine, such as a remote CI
+system.
 
 ## About this section
 
-This section will give a brief overview of Salvo's building, usage and results. If you want to dive into using Salvo to measure Envoy's performance change with an A/B testing, please read the detailed workflow documentation: [ENVOY_DEVELOP_WORKFLOW.md](./ENVOY_DEVELOP_WORKFLOW.md)
+This section will give a brief overview of Salvo's building, usage and results. If you want to dive
+into using Salvo to measure Envoy's performance change with an A/B testing, please read the detailed
+workflow documentation: [ENVOY_DEVELOP_WORKFLOW.md](./ENVOY_DEVELOP_WORKFLOW.md)
 
 ## Dependencies
 
-The [`install_deps.sh`](./install_deps.sh) script can be used to install any dependencies required by Salvo.
+The [`install_deps.sh`](./install_deps.sh) script can be used to install any dependencies required
+by Salvo.
 
 ## Building Salvo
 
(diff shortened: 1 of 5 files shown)


gyohuangxin:
@mum4k Can you review this? Thanks.

salvo/src/lib/test_source_manager.py
```diff
@@ -291,6 +291,66 @@ def test_find_all_images_from_specified_tags_fail(mock_source_tree):
...
+      scavenging_benchmark=True,
+  )
+
+  _generate_default_benchmark_images(job_control)
+
+  # Add an envoy image and specify additional versions to test
+  job_control.images.envoy_image="envoyproxy/envoy:v1.16.0"
+
+  # Specify test_single_image as true
+  job_control.images.test_single_image=True
```
mum4k, line 307: This looks mis-formatted, should we have spaces around the equal sign? This applies to multiple locations.

I know we had some auto formatting tools in this repository, could you try to find and execute them before we do the next review round?

mum4k (changes requested):
Thank you for the contribution @gyohuangxin!

salvo/README.md
```diff
@@ -94,7 +94,7 @@ images:
 
 `images.nighthawkBinaryImage`: Nighthawk tools will be sourced from this Docker image.
```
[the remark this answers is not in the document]
gyohuangxin: Good reminder, fixed.

salvo/api/image.proto
```diff
@@ -34,4 +34,10 @@ message DockerImages {
   // defined and benchmark the discovered image.
   // eg: [ "envoyproxy/envoy-dev:7da2adee8962c202999e96cb41e899deb34faf48",..]
   repeated string additional_envoy_images = 5;
+
+  // If this field is ture and additional_envoy_images is not defined, we will
```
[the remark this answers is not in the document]
gyohuangxin: Fixed.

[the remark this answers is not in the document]
gyohuangxin: | Can we explicitly say that those two are exclusive?

Yes, they are exclusive. If test_single_image=True and additional_envoy_images are defined, we will raise an error: 
```
      raise SourceManagerError(
          '"additional_envoy_image" cannot be set with "test_single_image" enabled')
```
The same applies to `test_single_commit` field.

salvo/api/source.proto
```diff
@@ -53,4 +53,10 @@ message SourceRepository {
   // A list of compiler options and flags to supply to bazel when building
   // the source of NightHawk or Envoy
   repeated BazelOption bazel_options = 7;
+
+  // If this field is ture and additional_hashes is not defined, we will
```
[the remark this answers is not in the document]
gyohuangxin: Fixed.

salvo/src/lib/test_source_manager.py
gyohuangxin: @mum4k I create another PR #134  to fix python files format, because it's large change which may confuse your review of this PR, can you merge it first? Then I will rebase this PR on that. Thanks!

salvo/src/lib/source_manager.py
```diff
@@ -183,12 +189,15 @@ def find_all_images_from_specified_sources(self) -> Set[str]:
     )
 
     # We have a source, see whether additional hashes are specified
-    addtional_hashes = source_repo.additional_hashes
-    if addtional_hashes:
-      hash_set = hash_set.union(addtional_hashes)
+    test_single_commit= source_repo.test_single_commit
```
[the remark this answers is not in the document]
mum4k: As you suggested below, let's address formatting in a separate PR.

salvo/src/lib/test_source_manager.py
mum4k: Thank you for making the review easier. Reviewed and merged. Please merge here and rerun the formatting tools on files modified in this PR.

salvo/api/image.proto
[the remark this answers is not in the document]
gyohuangxin: Yes, added it.

salvo/src/lib/test_source_manager.py
gyohuangxin: Done, please review again thanks!

salvo/api/image.proto
```diff
@@ -34,4 +34,11 @@ message DockerImages {
   // defined and benchmark the discovered image.
   // eg: [ "envoyproxy/envoy-dev:7da2adee8962c202999e96cb41e899deb34faf48",..]
   repeated string additional_envoy_images = 5;
+
+  // If this field is true and additional_envoy_images is not defined, we will
+  // only test envoy image which is defined in envoy_image field. It's useful
+  // for the scenario that someone want to test different configuration with 
+  // specific envoy image. test_single_image and additional_envoy_images are
+  // mutually exclusive and cannot be defined at the same time.
+  bool test_single_image = 6;
```
mum4k, line 43: Sorry, this didn't occur to me before. Since the fields are mutually exclusive, would it be better to define them as a oneof proto field?

https://developers.google.com/protocol-buffers/docs/proto3#oneof

Applies to the change in `source.proto` as well.

mum4k (changes requested):

salvo/api/image.proto
gyohuangxin: It's a nice suggestion and I've tried to define them as a `oneof` proto like:
```
  oneof test_mode {
    // Specify a set of additional hashes for comparing benchmark results.
    // This field is optional and if unspecified, we will default to using
    // one commit prior to the specified "commit_hash" field. For each
    // hash specfied we will run a benchmark using the Envoy binary built
    // from that source commit
    repeated string additional_hashes = 7;
    
    // If this field is true and additional_hashes is not defined, we will
    // only test envoy commit which is defined in commit_hash field. It's useful
    // for the scenario that someone want to test different configuration with 
    // specific envoy commit. test_single_commit and additional_hashes are
    // mutually exclusive and cannot be defined at the same time
    bool test_single_commit = 8;
  }
```

But the fileds in `oneof` cannot have label `repeated`: 
```
api/source.proto:52:5: Fields in oneofs must not have labels (required / optional / repeated).
```
If we delete repeated label, then it means `additional_hashes` cannot defined as a list, such as `additional_hashes: [02dff6e0235e2f497fb57eb0b7c07cc091c538df, b435d3a3baa39f1a15cd68625da085cfa16ae957]`.

mum4k: That is a good point, sorry for missing that. We typically go around that by wrapping the repeated field in a new message, but we shouldn't do that here, since that would be a breaking API change.

Let's proceed with what we have now. Thank you for exploring this option.

mum4k (approved):

PR #134, opened 2022-04-12
[Salvo] Fix python files format

gyohuangxin:


mum4k (approved):
Thank you, I have filed #135 to track addition of format checking into CI.

Until that is resolved, we should run these scripts manually on each PR.

PR #140, opened 2022-04-22
Add first AZP script

### LINKED ISSUE #139: [Salvo] Migrate CI workflows to AZP

gyohuangxin:
This PR fixes #139, adds the first AZP script to verify CI workflows can work well on AZP.

### DIFF
--- a/.azure-pipelines/bazel.yml
+++ b/.azure-pipelines/bazel.yml
@@ -1,0 +1,29 @@
+parameters:
+- name: ciTarget
+  displayName: "CI target"
+  type: string
+  default: build 
+
+steps:
+- bash: |
+    echo "disk space at beginning of build:"
+    df -h
+  displayName: "Check disk space at beginning"
+
+- script: ci/run_envoy_docker.sh 'ci/do_ci.sh ${{ parameters.ciTarget }}'
+  workingDirectory: $(Build.SourcesDirectory)
+  displayName: "Run the CI script"
+  env:
+    BAZEL_REMOTE_CACHE: $(LocalBuildCache)
+    AZP_BRANCH: $(Build.SourceBranch)
+    AZP_SHA1: $(Build.SourceVersion)
+    DOCKERHUB_USERNAME: $(DockerUsername)
+    DOCKERHUB_PASSWORD: $(DockerPassword)
+
+- bash: |
+    echo "disk space at end of build:"
+    df -h
+    # Cleanup offending files with unicode names
+    rm -rf $(Build.StagingDirectory)/tmp/*/*/external/go_sdk/test/fixedbugs
+  displayName: "Check disk space at end"
+  condition: always()
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -1,0 +1,60 @@
+trigger:
+  branches:
+    include:
+    - "main"
+
+stages:
+- stage: check
+  dependsOn: []
+  pool: "x64-large"
+  jobs:
+  - job: build_and_format
+    displayName: "do_ci.sh"
+    dependsOn: []
+    strategy:
+      maxParallel: 2
+      matrix:
+        build:
+          CI_TARGET: "build"
+        format:
+          CI_TARGET: "check_format"
+    timeoutInMinutes: 120
+    steps:
+    - template: bazel.yml
+      parameters:
+        ciTarget: $(CI_TARGET)
+
+- stage: test
+  dependsOn: ["check"]
+  pool: "x64-large"
+  jobs:
+  - job: test
+    displayName: "do_ci.sh"
+    strategy:
+      # Both test and benchmark need dedicated resources for stability.
+      maxParallel: 1
+      matrix:
+        test:
+          CI_TARGET: "test"
+    timeoutInMinutes: 120
+    steps:
+    - template: bazel.yml
+      parameters:
+        ciTarget: $(CI_TARGET)
+
+- stage: coverage
+  dependsOn: ["test"]
+  pool: "x64-large"
+  jobs:
+  - job: coverage
+    displayName: "do_ci.sh"
+    strategy:
+      maxParallel: 1
+      matrix:
+        coverage:
+          CI_TARGET: "coverage"
+    timeoutInMinutes: 120
+    steps:
+    - template: bazel.yml
+      parameters:
+        ciTarget: $(CI_TARGET)
--- a/ci/run_envoy_docker.sh
+++ b/ci/run_envoy_docker.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+
+set -e
+
+# shellcheck source=ci/envoy_build_sha.sh
+# . "$(dirname "$0")"/envoy_build_sha.sh
+
+function is_windows() {
+  [[ "$(uname -s)" == *NT* ]]
+}
+
+read -ra ENVOY_DOCKER_OPTIONS <<< "${ENVOY_DOCKER_OPTIONS:-}"
+
+# TODO(phlax): uppercase these env vars
+export HTTP_PROXY="${http_proxy:-}"
+export HTTPS_PROXY="${https_proxy:-}"
+export NO_PROXY="${no_proxy:-}"
+
+if is_windows; then
+  [[ -z "${IMAGE_NAME}" ]] && IMAGE_NAME="envoyproxy/envoy-build-windows2019"
+  # TODO(sunjayBhatia): Currently ENVOY_DOCKER_OPTIONS is ignored on Windows because
+  # CI sets it to a Linux-specific value. Undo this once https://github.com/envoyproxy/envoy/issues/13272
+  # is resolved.
+  ENVOY_DOCKER_OPTIONS=()
+  # Replace MSYS style drive letter (/c/) with Windows drive letter designation (C:/)
+  DEFAULT_ENVOY_DOCKER_BUILD_DIR=$(echo "${TEMP}" | sed -E "s#^/([a-zA-Z])/#\1:/#")/envoy-docker-build
+  BUILD_DIR_MOUNT_DEST=C:/build
+  SOURCE_DIR=$(echo "${PWD}" | sed -E "s#^/([a-zA-Z])/#\1:/#")
+  SOURCE_DIR_MOUNT_DEST=C:/source
+  START_COMMAND=("bash" "-c" "cd /c/source && export HOME=/c/build && $*")
+else
+  [[ -z "${IMAGE_NAME}" ]] && IMAGE_NAME="envoyproxy/envoy-build-ubuntu"
+  # We run as root and later drop permissions. This is required to setup the USER
+  # in useradd below, which is need for correct Python execution in the Docker
+  # environment.
+  ENVOY_DOCKER_OPTIONS+=(-u root:root)
+  ENVOY_DOCKER_OPTIONS+=(-v /var/run/docker.sock:/var/run/docker.sock)
+  ENVOY_DOCKER_OPTIONS+=(--cap-add SYS_PTRACE --cap-add NET_RAW --cap-add NET_ADMIN)
+  DEFAULT_ENVOY_DOCKER_BUILD_DIR=/tmp/envoy-docker-build
+  BUILD_DIR_MOUNT_DEST=/build
+  SOURCE_DIR="${PWD}"
+  SOURCE_DIR_MOUNT_DEST=/source
+  START_COMMAND=("/bin/bash" "-lc" "groupadd --gid $(id -g) -f envoygroup \
+    && useradd -o --uid $(id -u) --gid $(id -g) --no-create-home --home-dir /build envoybuild \
+    && usermod -a -G pcap envoybuild \
+    && chown envoybuild:envoygroup /build \
+    && sudo -EHs -u envoybuild bash -c 'cd /source && $*'")
+fi
+
+# The IMAGE_ID defaults to the CI hash but can be set to an arbitrary image ID (found with 'docker
+# images').
+[[ -z "${IMAGE_ID}" ]] && IMAGE_ID="${ENVOY_BUILD_SHA}"
+[[ -z "${ENVOY_DOCKER_BUILD_DIR}" ]] && ENVOY_DOCKER_BUILD_DIR="${DEFAULT_ENVOY_DOCKER_BUILD_DIR}"
+# Replace backslash with forward slash for Windows style paths
+ENVOY_DOCKER_BUILD_DIR="${ENVOY_DOCKER_BUILD_DIR//\\//}"
+mkdir -p "${ENVOY_DOCKER_BUILD_DIR}"
+
+[[ -t 1 ]] && ENVOY_DOCKER_OPTIONS+=("-it")
+[[ -f .git ]] && [[ ! -d .git ]] && ENVOY_DOCKER_OPTIONS+=(-v "$(git rev-parse --git-common-dir):$(git rev-parse --git-common-dir)")
+[[ -n "${SSH_AUTH_SOCK}" ]] && ENVOY_DOCKER_OPTIONS+=(-v "${SSH_AUTH_SOCK}:${SSH_AUTH_SOCK}" -e SSH_AUTH_SOCK)
+
+export ENVOY_BUILD_IMAGE="${IMAGE_NAME}:${IMAGE_ID}"
+
+time docker pull "${ENVOY_BUILD_IMAGE}"
+
+# Since we specify an explicit hash, docker-run will pull from the remote repo if missing.
+docker run --rm \
+       "${ENVOY_DOCKER_OPTIONS[@]}" \
+       -v "${ENVOY_DOCKER_BUILD_DIR}":"${BUILD_DIR_MOUNT_DEST}" \
+       -v "${SOURCE_DIR}":"${SOURCE_DIR_MOUNT_DEST}" \
+       -e AZP_BRANCH \
+       -e AZP_SHA1 `# unique` \
+       -e HTTP_PROXY \
+       -e HTTPS_PROXY \
+       -e NO_PROXY \
+       -e BAZEL_STARTUP_OPTIONS \
+       -e BAZEL_BUILD_EXTRA_OPTIONS \
+       -e BAZEL_EXTRA_TEST_OPTIONS \
+       -e BAZEL_REMOTE_CACHE \
+       -e ENVOY_STDLIB \
+       -e BUILD_REASON \
+       -e BAZEL_REMOTE_INSTANCE \
+       -e GCP_SERVICE_ACCOUNT_KEY \
+       -e NUM_CPUS \
+       -e ENVOY_RBE \
+       -e ENVOY_BUILD_IMAGE \
+       -e ENVOY_SRCDIR \
+       -e ENVOY_BUILD_TARGET \
+       -e ENVOY_BUILD_DEBUG_INFORMATION \
+       -e SYSTEM_PULLREQUEST_PULLREQUESTNUMBER \
+       -e GCS_ARTIFACT_BUCKET \
+       -e GITHUB_TOKEN \
+       -e BUILD_SOURCEBRANCHNAME \
+       -e BAZELISK_BASE_URL \
+       -e ENVOY_BUILD_ARCH \
+       -e SLACK_TOKEN \
+       -e BUILD_URI\
+       -e REPO_URI \
+       -e SYSTEM_STAGEDISPLAYNAME \
+       -e SYSTEM_JOBDISPLAYNAME \
+       -e SYSTEM_PULLREQUEST_PULLREQUESTNUMBER \
+       -e DOCKERHUB_USERNAME `# unique` \
+       -e DOCKERHUB_PASSWORD `# unique` \
+       "${ENVOY_BUILD_IMAGE}" \
+       "${START_COMMAND[@]}"


gyohuangxin:
@mum4k Any way to test this workflow in this PR? Or only we can test it after it merged?

.azure-pipelines/bazel.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- script: ci/run_envoy_docker.sh 'ci/do_ci.sh ${{ parameters.ciTarget }}'
+  workingDirectory: $(Build.SourcesDirectory)
+  displayName: "Run the CI script"
+  env:
+    BAZEL_REMOTE_CACHE: $(LocalBuildCache)
+    AZP_BRANCH: $(Build.SourceBranch)
+    AZP_SHA1: $(Build.SourceVersion)
+    DOCKERHUB_USERNAME: $(DockerUsername)
+    DOCKERHUB_PASSWORD: $(DockerPassword)
```
gyohuangxin, line 21: Do we need to add above environment variables in repo settings? Can you help me add them, because I don't have the permission?

mum4k: These variables will be set upstream in AZP.

I suspect that we don't need most of them. Can we remove all and only keep the `BAZEL_REMOTE_CACHE` ?

.azure-pipelines/pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+  pool: "x64-large"
+  jobs:
+  - job: build_and_format
+    displayName: "do_ci.sh"
+    dependsOn: []
+    strategy:
+      maxParallel: 2
+      matrix:
+        build:
+          CI_TARGET: "build"
```
mum4k, line 18: Correct me if I am wrong, but I suspect Salvo doesn't need a lot of resources to run the `build`, `check_format`, `test` and `coverage` stages. The way this is set up now, each one of them will run on a separate VM.

Can we define them all in a single stage running them as parallel jobs? Same as the build` and `check_format` here.

ci/run_envoy_docker.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
mum4k, line 1: If we are cloning this file, we should clone it from the original in the Envoy repo:

https://github.com/envoyproxy/envoy/blob/main/ci/run_envoy_docker.sh

We should also document the maintenance procedure in a `MAINTAINERS.md` file indicating that this file needs to be cloned regularly, since Envoy changes it from time to time.

Example from Nighthawk:
https://github.com/envoyproxy/nighthawk/blob/main/MAINTAINERS.md

.azure-pipelines/bazel.yml
```diff
@@ -0,0 +1,29 @@
...
+  type: string
+  default: build 
+
+steps:
+- bash: |
+    echo "disk space at beginning of build:"
+    df -h
+  displayName: "Check disk space at beginning"
+
+- script: ci/run_envoy_docker.sh 'ci/do_ci.sh ${{ parameters.ciTarget }}'
```
mum4k, line 13: Do we know for a fact that we need the `ci/run_envoy_docker.sh`? It would be easier if we could make it work without that dependency, since it will require regular updates / maintenance.

Maybe we could first try without and add it here if we cannot make the CI work without?

mum4k (changes requested):
We can start testing AZP at this PR before we merge it in. Let's address the review comments first. I can then define the pipeline in AZP and this will start executing it on every new commit. That is how you will be able to iterate and debug.

.azure-pipelines/bazel.yml
gyohuangxin: For sure, I will remove them.

.azure-pipelines/pipelines.yml
gyohuangxin: Yes, I agreed. I will move them into one single stage.

.azure-pipelines/bazel.yml
gyohuangxin: Yes, got it. Sorry I just copied from Nighthawk's AZP configuration. But if we want to run without `ci/run_envoy_docker.sh`, we also need to find a way to make CI workflow run in a correct Envoy docker container, like https://github.com/envoyproxy/envoy-perf/blob/main/.circleci/config.yml#L3. 
At first, we can draft a simple script to implement it. And if we need more features in `ci/run_envoy_docker.sh` really, we can use it again. Does it make sense?

mum4k: Thinking more, it might be simplest if we just copy and use the same script. We will need the container for the tooling that it already contains. And if we use the same script in all three repos, it will be easier to debug and to understand the CI setup.

Going to resolve this conversation. We can continue discussing how to document the maintenance procedure in the other comment thread.

gyohuangxin:
@mum4k Thanks for reviewing and giving advises, I updated following your comments, and kept the ci/run_envoy_dcocker.sh the same as Envoy's and Nighthawk's, so I also need to add `.bazelrc` and `ci/envoy_build_sha.sh`. I'm testing if it can work and we can discuss the maintenance procedure here.

.azure-pipelines/pipelines.yml
```diff
@@ -0,0 +1,29 @@
...
+stages:
+- stage: salvo_test_and_check
+  dependsOn: []
+  pool: "x64-large"
+  jobs:
+  - job: salvo_test_and_check
+    displayName: "do_ci.sh"
+    dependsOn: []
+    strategy:
+      maxParallel: 4
```
mum4k, line 15: Sorry, when configured like this, the parallel jobs actually request 4 VMs and execute in parallel on them. That seems like an unnecessary amount of resources for Salvo build / check / test steps.

https://dev.azure.com/cncf/envoy/_build/results?buildId=106913&view=results

We will want to have dedicated VMs for performance tests, but we can save some resources here. Can we refactor this so that these jobs either run serially on 1 VM at a time, or one after each other on the same VM?

mum4k (changes requested):
(fyi) I created the pipeline upstream on AZP and it should now execute on every commit:

https://dev.azure.com/cncf/envoy/_build?definitionId=53&_a=summary

You should be able to see the results when you create a Microsoft account.

Let's work on tweaking the amount of resources this pipeline uses.

.azure-pipelines/pipelines.yml
gyohuangxin: @mum4k Correct me I was wrong, I failed to find a way in AZP to run above parallel 4 steps in one VM unless we add one new `CI_TARGET` like 'all' which will execute all steps. What do you think?

mum4k: I think there are two ways to achieve this:

1. Define these as 4 jobs that depend on each other, i.e. the next job only starts after the first one finishes. See the `dependsOn` list.
1. What you mentioned - add one CI target that executes all the steps in one script.

Approach (1) will take longer, because it will set up a VM for each of the jobs, but it will also execute hermetically and will only use one VM at a time. If it won't take too long to run it this way, I think we should prefer (1). Can you give it a try?

mum4k: (fyi) Configuring (1) might be as simple as setting `maxParallel: 1`.

gyohuangxin: > Define these as 4 jobs that depend on each other, i.e. the next job only starts after the first one finishes. See the dependsOn list.

Correct me if it's wrong. I tried defining them depend on each other like this: 
```
    strategy:
      maxParallel: 1
      matrix:
        build:
          CI_TARGET: "build"
        format:
          dependsOn: build
          CI_TARGET: "check_format"
        test:
          dependsOn: test
          CI_TARGET: "test"
        coverage:
          dependsOn: coverage
          CI_TARGET: "coverage"
```
however, it seems that it still created 4 VMs, you can find those agent machines' names here:
https://dev.azure.com/cncf/envoy/_build/results?buildId=106923&view=logs&j=1624a51a-a39c-5947-4081-f6b573564ebe&t=f59bbc51-ac17-4b90-a598-d95b3f333e77&l=3
https://dev.azure.com/cncf/envoy/_build/results?buildId=106923&view=logs&j=27dc7755-952e-5a2b-a136-5632fca0d844&t=2dae8299-bfbc-417e-8ecc-06c2aec77099&l=3
https://dev.azure.com/cncf/envoy/_build/results?buildId=106923&view=logs&j=45e0cd60-018b-5371-3188-0d384cbaa76f&t=ac498b66-e74e-472f-a607-2988e9c239ea&l=3
https://dev.azure.com/cncf/envoy/_build/results?buildId=106923&view=logs&j=35882957-0705-54d1-237e-39e046a21a26&t=6741b2c4-c2af-40c2-9239-5bdec56e2c57&l=3

gyohuangxin: Sorry for misunderstanding what you said, yes, though approach (1) will create 4 VMs, it will use one VM at the same time. And we the simplest way to implement it is setting `maxParallel: 1`

gyohuangxin:
@mum4k The dependencies of Salvo must be installed with root users: https://github.com/envoyproxy/envoy-perf/blob/main/salvo/install_deps.sh#L3,
https://dev.azure.com/cncf/envoy/_build/results?buildId=106929&view=logs&j=45e0cd60-018b-5371-3188-0d384cbaa76f&t=e607ad86-6e5a-56be-ba12-2a2ce883a001&l=70
However, `run_envoy_docker.sh` will use an `envoybuild` user, so it will fail.
Is it allowed to change `run_envoy_docker.sh` to use root user, or add `envoybuild` to sudoers group?

gyohuangxin:
@mum4k The dependencies is that we need update and install these ubuntu packages to perform the python tests, https://github.com/envoyproxy/envoy-perf/blob/main/salvo/install_deps.sh#L15-L20, which is required root users.
And the envoy image we are using is `envoyproxy/envoy-build-ubuntu:e7ea4e81bbd5028abb9d3a2f2c0afe063d9b62c0`, therefore we can add above packages based on the image if possible.

mum4k:
Sorry @gyohuangxin, what I meant is - can you identify **exactly** which package is missing when compared to the Envoy's docker image?

Note, the Envoy docker image is built from this script:
- https://github.com/envoyproxy/envoy-build-tools/blob/main/build_container/build_container_ubuntu.sh

Once we identify exactly which package is missing, we can either find a way to run without the package, or talk to the Envoy team and see if we can add the package.

mum4k:
Looking closer, it looks like the Envoy docker image is missing the following, please confirm before we proceed:

- docker.io
- python3-pytest
- python3-docker

Looks like it already has `openjdk-11-jdk`.

Let's try to reach out to the maintainers of the https://github.com/envoyproxy/envoy-build-tools repository and see if they would accept addition of these packages on the bases that Salvo is aiming to serve the Envoy ecosystem. We can develop our own docker image in the future, but it would help us we we can avoid that investment for now.

Please mention me on any issue / PR related to this.

gyohuangxin:
@mum4k Thanks for the information, I will double confirm that. I will try to build a image from https://github.com/envoyproxy/envoy-build-tools/blob/main/build_container/Dockerfile-ubuntu, then I will check which packages are missing till I can build/test Salvo.  And then I will create an issue to discuss it in https://github.com/envoyproxy/envoy-build-tools repository and mention you.

mum4k:
Perfect, thank you @gyohuangxin!

gyohuangxin:
> Looking closer, it looks like the Envoy docker image is missing the following, please confirm before we proceed:
> 
> docker.io
> python3-pytest
> python3-docker
> Looks like it already has openjdk-11-jdk.

@mum4k  Could you tell how did you figure out that it already has openjdk-11-jdk, I build a image from above https://github.com/envoyproxy/envoy-build-tools/blob/main/build_container/Dockerfile-ubuntu but can't find it in listed apt package:
```
Listing... Done
....
ninja-build/now 1.8.2-1 amd64 [installed,local]
openssh-client/now 1:7.6p1-4ubuntu0.6 amd64 [installed,local]
openssl/now 1.1.1-1ubuntu2.1~18.04.15 amd64 [installed,local]
passwd/now 1:4.5-1ubuntu2.2 amd64 [installed,local]
patch/now 2.7.6-2ubuntu1.1 amd64 [installed,local]
perl/now 5.26.1-6ubuntu0.5 amd64 [installed,local]
perl-base/now 5.26.1-6ubuntu0.5 amd64 [installed,local]
perl-modules-5.26/now 5.26.1-6ubuntu0.5 all [installed,local]
pinentry-curses/now 1.1.0-1 amd64 [installed,local]
procps/now 2:3.3.12-3ubuntu1.2 amd64 [installed,local]
python/now 2.7.15~rc1-1 amd64 [installed,local]
python-apt-common/now 1.6.5ubuntu0.7 all [installed,local]
python-minimal/now 2.7.15~rc1-1 amd64 [installed,local]
python-pip/now 9.0.1-2.3~ubuntu1.18.04.5 all [installed,local]
python-pip-whl/now 9.0.1-2.3~ubuntu1.18.04.5 all [installed,local]
python-pkg-resources/now 39.0.1-2 all [installed,local]
python-setuptools/now 39.0.1-2 all [installed,local]
python2.7/now 2.7.17-1~18.04ubuntu1.7 amd64 [installed,local]
python2.7-minimal/now 2.7.17-1~18.04ubuntu1.7 amd64 [installed,local]
python3/now 3.6.7-1~18.04 amd64 [installed,local]
python3-apt/now 1.6.5ubuntu0.7 amd64 [installed,local]
python3-dbus/now 1.2.6-1 amd64 [installed,local]
python3-gi/now 3.26.1-2ubuntu1 amd64 [installed,local]
python3-minimal/now 3.6.7-1~18.04 amd64 [installed,local]
python3-software-properties/now 0.96.24.32.18 all [installed,local]
python3.10/now 3.10.4-1+bionic2 amd64 [installed,local]
python3.10-dev/now 3.10.4-1+bionic2 amd64 [installed,local]
python3.10-distutils/now 3.10.4-1+bionic2 all [installed,local]
python3.10-lib2to3/now 3.10.4-1+bionic2 all [installed,local]
python3.10-minimal/now 3.10.4-1+bionic2 amd64 [installed,local]
python3.6/now 3.6.9-1~18.04ubuntu1.7 amd64 [installed,local]
python3.6-minimal/now 3.6.9-1~18.04ubuntu1.7 amd64 [installed,local]
....

**Comment by mum4k** [comment]
@gyohuangxin you are probably right, I saw it here:

https://github.com/envoyproxy/envoy-build-tools/blob/cc97d26baf03e87982c6aaad6eccc6411543cf4c/build_container/build_container_ubuntu.sh#L127

But I see now that this only gets installed for the `aarch64` architecture.

**Comment by gyohuangxin** [comment]
@mum4k Fine, I just wanted to confirm. Thanks.

**Comment by gyohuangxin** [comment]
@mum4k As discussed in https://github.com/envoyproxy/envoy-build-tools/issues/174, I finally can install above dependencies  without using apt. 
- `docker.io` is not necessary, because `docker-ce-cli` is installed and works.
- `python3-pytest`, `python3-docker` can be installed with pip.
- `openjdk-11-jdk` isn't needed to be installed in image, using remote jdk with bazel `--java_runtime_version=remotejdk_11` works. Thanks to Lizan's comment: https://github.com/envoyproxy/envoy-build-tools/issues/174#issuecomment-1121426076

You can check all CI workflows on AZP worked well and passed: https://dev.azure.com/cncf/envoy/_build/results?buildId=108085&view=logs&j=1624a51a-a39c-5947-4081-f6b573564ebe

Therefore, we don't need to add any packages in envoy image at this stage, we can close https://github.com/envoyproxy/envoy-build-tools/issues/174 and reopen it if we meet packages issues in the future. WDYT?

File: .bazelrc [hunk 6]
```diff
@@ -0,0 +1,410 @@
+# The following .bazelrc content is forked from the main Envoy repository.                  # unique
```
mum4k, line 1: Looks like this file was actually cloned from Nighthawk, not from Envoy. That is fine to do, since we need to be able to build Nighthawk, but we should update the commentary in this file.

ci/run_envoy_docker.sh
mum4k: Looks like the PR clones a few files from a couple of locations. When updating the MAINTAINERS.md, please list all the files we cloned and their original locations, so we know how to update this when necessary.

salvo/install_deps.sh
```diff
@@ -3,7 +3,6 @@
 if [ ${UID} -ne 0 ]
 then
   echo "This script needs root priviliges to install dependencies. Continuing may elicit failures from tests"
```
mum4k, line 5: This script isn't just used in CI, I believe it is also used as the script to install Salvo dependencies locally when starting to develop Salvo:

https://github.com/envoyproxy/envoy-perf/tree/main/salvo#dependencies

Will the script still work for that purpose after the changes here? Or should we have a separate script for CI?

mum4k (changes requested):
Great news on the Envoy docker image, thank you for investigating this @gyohuangxin.

.bazelrc
gyohuangxin: @mum4k Sure, I'll chang Envoy to Nighthawk.

salvo/install_deps.sh
gyohuangxin: @mum4k You're right. It may not work locally on the machine has not `docker.io` or `docker-ce-cli`. I think a separate CI script is a better option. I will create another script named `install_deps_ci.sh` and keep this scripts unchanged.

gyohuangxin:
> Looks like the PR clones a few files from a couple of locations. When updating the MAINTAINERS.md, please list all the files we cloned and their original locations, so we know how to update this when necessary.

1. [ci/run_envoy_docker.sh](https://github.com/envoyproxy/envoy-perf/pull/140/files#diff-2e2b8c5b195a96a1c8f999beda3ea91d5748ab5bb4a097d3fb711830016bdb19) cloned from https://github.com/envoyproxy/envoy/blob/main/ci/run_envoy_docker.sh
2. [ci/envoy_build_sha.sh](https://github.com/envoyproxy/envoy-perf/pull/140/files#diff-6d0079005e8f336dd97cca7070e0a8e2ac840a0a72a78140bb8bc2326fc0e939) cloned from https://github.com/envoyproxy/envoy/blob/main/ci/envoy_build_sha.sh
3. [.bazelrc](https://github.com/envoyproxy/envoy-perf/pull/140/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832f) cloned from https://github.com/envoyproxy/nighthawk/blob/main/.bazelrc
4. [.azure-pipelines/bazel.yml](https://github.com/envoyproxy/envoy-perf/pull/140/files#diff-634d3872fb352a01a0a3780a6c242e0c20f1df44c34efedf90c2f1fa3697bc49) cloned from https://github.com/envoyproxy/nighthawk/blob/main/.azure-pipelines/bazel.yml
5. [.azure-pipelines/pipelines.yml](https://github.com/envoyproxy/envoy-perf/pull/140/files#diff-8ca122481a2ae193003b523bf5b7a9cfa0e914be502d483b6adc1d06668c724a) cloned from https://github.com/envoyproxy/nighthawk/blob/main/.azure-pipelines/pipelines.yml

No code changes in file 1 and file 2, but changes in 3, 4, 5.

mum4k:
Thank you for listing the cloned files. I think it is ok for us to have our own copy of `.azure-pipelines/bazel.yml` and `.azure-pipelines/pipelines.yml`. These are likely to evolve over time as we add more tests and we don't need to worry about their origin or about keeping them in sync.

For the other three files, please create a file `MAINTAINERS.md` and document a "maintenance procedure", e.g. how to update these files in the future. Something like what Nighthawk has here:

https://github.com/envoyproxy/nighthawk/blob/main/MAINTAINERS.md#updates-to-the-envoy-dependency

We need the steps that talk about syncing (copying) files form their originals. This is a procedure we should start executing regularly to ensure we are following Envoy as they update the docker image and our CI remains functional.

gyohuangxin:
@mum4k Thanks for this guide. I went through Nighthawk's `maintainers.md` and got confused with:
1. Do we also need copy bazel/repositories.bzl and .bazelversion
2. There was one different about image in Salvo's .bazelrc: https://github.com/envoyproxy/envoy-perf/pull/140/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832fR286. So we need marked it with comment `# unique`? Then delete these lines with `#unique` because it's related to Nighthawk not Salvo?https://github.com/envoyproxy/envoy-perf/blob/20c834e57ffb43e56b30f412c9321ee90d4cf044/.bazelrc#L75-L78

mum4k:
@gyohuangxin we don't need to worry about `bazel/repositories.bzl`, because that is Nigthhawk specific. Nighthawk imports Envoy internals and this is the hash at which the import happens.

We only need `.bazelversion` if we decide to tell our users that they should build Salvo with [bazelisk](https://github.com/bazelbuild/bazelisk). That is probably a good idea, since it is what both Envoy and Nighthawk do, but we don't need to address that in this PR. If you choose to do this I would prefer if we send a new PR to introduce `.bazelversion` with the relevant documentation updates.

Yes I would recommend marking lines that need to be unique for Salvo. If you are choosing to copy .bazelrc from Nighthawk, we should keep everything as is in Nighthawk and just add a new marker for lines unique to us, say `# Salvo unique`?

gyohuangxin:
Opps, if we add `# Salvo unique` after https://github.com/gyohuangxin/envoy-perf/blob/788cb4335e09277b4bce8aa8bba59b905afa1ac4/.bazelrc#L286, the image name and tags will be like `envoyproxy/envoy-build-ubuntu:e7ea4e81bbd5028abb9d3a2f2c0afe063d9b62c0 # Salvo unique`, which will make CI fail.

https://dev.azure.com/cncf/envoy/_build/results?buildId=108393&view=logs&j=27dc7755-952e-5a2b-a136-5632fca0d844&t=3c53ca42-53a4-5ff7-2c8b-2440b494f5f9&l=14

mum4k:
I think that problem is specific to this line, because it is being parsed [here](https://github.com/envoyproxy/envoy-perf/blob/788cb4335e09277b4bce8aa8bba59b905afa1ac4/ci/envoy_build_sha.sh) and the parsing likely doesn't correctly exclude the comment at the end. We could work around it by placing the comment above the line.

However we should try to minimize the number of lines we keep unique, since that diverges us from the rest of the ecosystem and adds maintenance tasks. Can you help me understand why we can't use the same image as used at the source?

gyohuangxin:
The reason why we use this image is that the Salvo cannot be tested and built successfully in the recent `envoyproxy/envoy-build-ubuntu` image. The new image bases on python3.10 but Salvo works on python3.6, so we should update packages and fix many dependencies issues which will take a lot of time. I think it's better to keep using old image and then update to python3.10 in another PR. WDYT?

> I think that problem is specific to this line, because it is being parsed [here](https://github.com/envoyproxy/envoy-perf/blob/788cb4335e09277b4bce8aa8bba59b905afa1ac4/ci/envoy_build_sha.sh) and the parsing likely doesn't correctly exclude the comment at the end. We could work around it by placing the comment above the line.

Yes, it's the cause.  It's a good idea to place the comment above the line.

mum4k:
Thank you for explaining the reasoning @gyohuangxin, postponing the update of the image to the next PR makes sense to me and is preferred. Please consider opening an issue and linking the issue from the comment in the file where we include the `# Salvo unique` marker.

MAINTAINERS.md
```diff
@@ -0,0 +1,32 @@
...
+1. Sync (copy) [ci/envoy_build_sha.sh](ci/envoy_build_sha.sh) from
+   [Envoy's version](https://github.com/envoyproxy/envoy/blob/main/ci/envoy_build_sha.sh).
+   Be sure to retain our local modifications, all lines that are unique to
+   Salvo are marked with comment `# Salvo unique`.
+1. Run `ci/do_ci.sh test`. Sometimes the dependency update comes with changes
+   that break our build. Include any changes required to Salvo to fix that
+   in the same PR.
+1. If the PR ends up modifying any python files, execute `ci/do_ci.sh fix_format`
+   to reformat the files and avoid a CI failure.
+1. Create a PR with a title like `Update Envoy to 9753819 (Jan 24th 2021)`,
```
mum4k, line 31: This PR title probably doesn't fit Salvo well. In Nighthawk we are updating the Envoy dependency, so it works there. However here we are syncing a couple of files, but not updating anything to a particular commit. Also it isn't clear what the meaning of the number `9753819` is. Consider:

`Syncing configuration files from Nighthawk and Envoy (..date here..)`

MAINTAINERS.md
```diff
@@ -0,0 +1,32 @@
...
+## Updates to the Envoy and Nighthawk dependency for Salvo
+
+Salvo reuses large parts of Envoy's and Nighthawk's build system and codebase, so keeping Salvo up
+to date with Envoy's and Nighthawk's changes is an important maintenance task. When performing the
+update, follow this procedure:
+
+1. Create a fork of Salvo, or fetch upstream and merge changes into your fork if you already have
+one.
+1. Create a new branch from `main`, e.g. `envoy-update`.
+1. Sync (copy) [.bazelrc](.bazelrc) from
```
mum4k, line 14: We have a potential problem here - we are syncing configuration files from two repositories. If we copy from both of those at head (the latest commit) they may not be in sync with each other.

Ideally we would:
1. Copy files we need from Nighthawk at head (latest commit).
2. Look at Nighthawks configuration and determine the Envoy commit Nighthawk is updated to, i.e. y looking [here](https://github.com/envoyproxy/nighthawk/blob/f911358d0e453180b3cc0cd8affe4b3319cfe6f9/bazel/repositories.bzl#L3).
3. Copy the files from Envoy at the commit from (2).

salvo/install_deps_ci.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
mum4k, line 2: Should we add `set -eo pipefail`?

mum4k (changes requested):
Thank you for adding documentation for the maintenance procedure. Leaving a few notes.

MAINTAINERS.md
gyohuangxin: I can understand what you worried about, we copy `.bashrc` from Nighthawk's one while copy `ci/run_envoy_docker.sh` and `ci/envoy_build_sha.sh` from Envoy's ones. However, as you suggested, if we look at Nighthawk's `repositories.bzl` to find out the Envoy commit, then copy the files from Envoy at this commit finally, we will lose Nighthawk's unique parts. 

So how about this following guide:
1. Look at Nighthawks configuration and determine the Envoy commit Nighthawk is updated to, i.e. y looking [here](https://github.com/envoyproxy/nighthawk/blob/f911358d0e453180b3cc0cd8affe4b3319cfe6f9/bazel/repositories.bzl#L3).
2. Copy the files from Envoy at the commit from (1).
3. Be sure to retain Nighthawk's modifications in `.bazelrc` from [here](https://github.com/envoyproxy/nighthawk/blob/f911358d0e453180b3cc0cd8affe4b3319cfe6f9/.bazelrc) all lines that are unique to Nighthawk are marked with comment `# unique`.
4. Be sure to retain Salvo's modification from [.bazelrc](.bazelrc), all lines that are unique to
   Salvo are marked with comment `# Salvo unique`.

gyohuangxin: Maybe it's complicated for maintainers, I'm not sure if it's ok we sync files only from Envoy or only from Nighthawk? To be honest, I don't know much about their relationship and possible impact at the moment. Please tell me if I'm wrong.

salvo/install_deps_ci.sh
gyohuangxin: Yes, we should add `set -eo pipefail` in all CI scripts. Updated, thanks.

MAINTAINERS.md
mum4k: I think we should try to simplify this for files where we need to retain both Nighthawk's and Salvo's local modifications by copying them from Nighthawk instead of from Envoy. Looks like that only applies to `.bazelrc`. I.e. we can keep most of this as suggested in your last comment, but I would indicate to copy `.bazelrc` from Nighthawk.

gyohuangxin: @mum4k  I updated it about the `.bazelrc` synchronization, I tried to keep it as simple as possible. Is it ok?

MAINTAINERS.md
gyohuangxin: Thanks, updated.

MAINTAINERS.md
gyohuangxin: Well, the reason is to resolve the problem you mentioned above: 
```
We have a potential problem here - we are syncing configuration files from two repositories. If we copy from both of those at head (the latest commit) they may not be in sync with each other.
```
My understanding is that if  we copy from `.bazelrc` from Nighthawk's one, and copy `ci/run_envoy_docker.sh` and `ci/envoy_build_sha.sh` from Envoy's ones. The `.bazelrc` file may not sync with `ci/run_envoy_docker.sh` and `ci/envoy_build_sha.sh`. So we need to find a way to confirm the `.bazelrc` file copied from Nighthawk is updated to Envoy's `.bazelrc`. 

On second thought, this is no longer a problem, if we look at "Envoy commit Nighthawk update to", they are always in sync. Sorry for making simple things complicated, I will update again to keep it simple.

gyohuangxin: Updated: https://github.com/envoyproxy/envoy-perf/pull/140/commits/bd036c894e03bcda1aea991538f36d6cf6c2d5e2, please review again. Thanks.

PR #141, opened 2022-04-29
Add error trapping in Salvo test coverage checks

gyohuangxin:
The Salvo test and coverage CI workflows show green though there are errors: 
https://app.circleci.com/pipelines/github/gyohuangxin/envoy-perf/13/workflows/e1eb2506-0730-4012-9616-1fbbc9140927/jobs/47?invite=true#step-102-1050
https://app.circleci.com/pipelines/github/gyohuangxin/envoy-perf/13/workflows/e1eb2506-0730-4012-9616-1fbbc9140927/jobs/48?invite=true#step-102-1022

This PR aims to add trapping at above errors and fix them.

### DIFF
--- a/salvo/tools/coverage.sh
+++ b/salvo/tools/coverage.sh
@@ -36,6 +36,11 @@ CMD="${CMD} -o coverage/coverage.dat"
 
 # Extract all coverage data for salvo project files
 lcov -e coverage/coverage.dat -o coverage/salvo.dat '*/salvo/src/lib/*.py'
+if [ "$?" != "0" ]; then
+  echo "Failed to run test coverage checks."
+  exit 1
+fi
+
 
 # Redirect source file paths from the sandbox to the real source files.
 # Changes strings like this:
@@ -47,6 +52,10 @@ sed -e 's/SF.*\.runfiles\/salvo\/\(.*\)$/SF:\1/' -i coverage/salvo.dat
 # Generate HTML coverage report.
 mkdir -p coverage/html
 genhtml coverage/salvo.dat -o coverage/html
+if [ "$?" != "0" ]; then
+  echo "Failed to run test coverage checks."
+  exit 1
+fi
 echo "HTML coverage report generated, view by running:"
 echo "  (cd coverage/html && python3 -m http.server)"
 
--- a/salvo/tools/coverage.sh
+++ b/salvo/tools/coverage.sh
@@ -36,11 +36,6 @@ CMD="${CMD} -o coverage/coverage.dat"
 
 # Extract all coverage data for salvo project files
 lcov -e coverage/coverage.dat -o coverage/salvo.dat '*/salvo/src/lib/*.py'
-if [ "$?" != "0" ]; then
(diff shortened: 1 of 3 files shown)


salvo/tools/coverage.sh
```diff
@@ -35,6 +35,11 @@ CMD="${CMD} -o coverage/coverage.dat"
 
 # Extract all coverage data for salvo project files
 lcov -e coverage/coverage.dat -o coverage/salvo.dat '*/salvo/src/lib/*.py'
+if [ "$?" != "0" ]; then
```
mum4k, line 38: Checking the error after individual commands may still be error prone (we may forget to add checks when we add new commands).

Can we do it at the top of the file instead? This file could start as:

```
#!/bin/bash

# Exit immediately if a command exits with a non-zero status.
# pipefail indicates that the return value of a pipeline is the status
# of the last command to exit with a non-zero status.
set -eo pipefail
```

We may want to adopt this approach in all our CI scripts.

mum4k (changes requested):
Thank you for spotting this problem.

salvo/tools/coverage.sh
gyohuangxin: @mum4k Thanks for pointing out this better approach.

gyohuangxin:
@mum4k It always fails with the error: https://app.circleci.com/pipelines/github/gyohuangxin/envoy-perf/18/workflows/1a177034-18ed-43f1-8094-ed51d7c45c28/jobs/65?invite=true#step-102-1006, and I think it related to OOM. Do you have any idea about how to resolve it so that I can verify code changes?

mum4k:
@gyohuangxin running out of memory is the primary reason why we are switching away from CircleCI, we have more resources on AZP.

If memory is the bottleneck here, we may not be able to get this to work on CircleCI, or it might be throw-away work. Should we focus on the AZP migration first? We can disable portions of the CI script that cannot pass on CircleCI and then re-enable them once we are on AZP.

gyohuangxin:
@mum4k Sorry for the late reply due to holiday. Sure, I will focus on the migration first.

gyohuangxin:
@mum4k  Do you know how to invoke AZP retest without a new commit?

mum4k:
@gyohuangxin repokkiteh is currently configured to [retest via CircleCI](https://github.com/envoyproxy/envoy-perf/blob/ad15bca77f7f450705981081bc194de175716dd6/repokitteh.star#L6). We will have to switch it to AZP, see this [example from Nighthawk](https://github.com/envoyproxy/nighthawk/blob/f911358d0e453180b3cc0cd8affe4b3319cfe6f9/repokitteh.star#L4-L6).

Once we do that, I will also need to define the secret token for repokkiteh to use when calling the AZP APIs. Personally I would keep this to the end and retry by adding commits for now.

gyohuangxin:
@mum4k Thanks, understood.

gyohuangxin:
@mum4k Ready for review.

mum4k (approved):

PR #144, opened 2022-05-29
Update Salvo requirements to python3.10

### LINKED ISSUE #142: [Salvo] Migrate from Python 3.6 to Python3.10
Salvo bases on Python3.6 currently, but the latest `envoyproxy/envoy-build-ubuntu` image bases on Python3.10, so we still use a old image here: https://github.com/envoyproxy/envoy-perf/blob/main/.bazelrc#L286-L287
We should migrate Salvo to Python3.10, so that we can use the latest image.

gyohuangxin:
This PR fixed #142:
- Change the envoyproxy/envoy-build-ubuntu image to be the same as Nighthawk's and  Envoy's one.
- Update Salvo requirements to python3.10 and fix dependencies issues.
- Change unit tests threshold to 97%.


gyohuangxin:
@mum4k This PR is ready for review.

salvo/requirements.txt
```diff
@@ -1,35 +1,48 @@
...
-PyYAML==5.4
-requests==2.25.0
-s3transfer==0.3.3
-six==1.15.0
-typed-ast==1.4.1
-urllib3==1.26.5
-virtualenv==20.0.31
-websocket-client==0.44.0
-zipp==3.1.0
+attrs==21.4.0
```
mum4k, line 2: I know we did this before, however fixing requirements to a specific version (and on top of that to a specific patch version) will cause higher than usual amount of work in the future. I.e. there will be a large amount of changes we will have to do all at once.

What would it take to use `>=` instead of `==` ideally in all our dependencies? This will break CI when the dependency makes an incompatible change, but at least we will know about it soon. It should be part of our maintenance procedure to update the dependencies regularly.

salvo/tools/typecheck.sh
```diff
@@ -26,4 +26,4 @@ fi
 
 echo $PWD
 
-${PYTYPE} src -P bazel-bin:.
+${PYTYPE} src -P bazel-bin:. --disable=pyi-error
```
mum4k, line 29: Can we add a comment explaining what are we disabling here and why?

mum4k (changes requested):

salvo/tools/typecheck.sh
gyohuangxin: It will failed with below error without disabling this:
```
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/type_pb2.py:
external/com_google_protobuf/python: warning: directory does not exist.
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/empty_pb2.py:
external/com_google_protobuf/python: warning: directory does not exist.
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/descriptor_pb2.py:
external/com_google_protobuf/python: warning: directory does not exist.
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/wrappers_pb2.py:
external/com_google_protobuf/python: warning: directory does not exist.
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/api_pb2.py:
external/com_google_protobuf/python: warning: directory does not exist.
INFO: Elapsed time: 33.397s, Critical Path: 11.85s
INFO: 254 processes: 56 internal, 198 processwrapper-sandbox.
INFO: Build completed successfully, 254 total actions
++ which pytype
+ PYTYPE=
+ true
+ '[' -z '' -a -f /root/.local/bin/pytype ']'
+ PYTYPE=/root/.local/bin/pytype
+ '[' -z /root/.local/bin/pytype ']'
+ echo /envoy-perf/salvo
/envoy-perf/salvo
+ /root/.local/bin/pytype src -P bazel-bin:.
Computing dependencies
Analyzing 35 sources with 5 local dependencies
ninja: Entering directory `.pytype'
[7/40] check src.lib.source_tree
FAILED: /envoy-perf/salvo/.pytype/pyi/src/lib/source_tree.pyi
/usr/bin/python3.10 -m pytype.single --imports_info /envoy-perf/salvo/.pytype/imports/src.lib.source_tree.imports --module-name src.lib.source_tree --platform linux -V 3.10 -o /envoy-perf/salvo/.pytype/pyi/src/lib/source_tree.pyi --analyze-annotated --enable-nested-classes --nofail --quick /envoy-perf/salvo/src/lib/source_tree.py
File "/envoy-perf/salvo/src/lib/source_tree.py", line 12, in <module>: Couldn't import pyi for 'api.source_pb2' [pyi-error]
  Can't find pyi for 'global___FieldDescriptorProto.Label', referenced from 'google.protobuf.descriptor_pb2'

For more details, see https://google.github.io/pytype/errors.html#pyi-error
ninja: build stopped: subcommand failed.
Leaving directory '.pytype'
```

It appears to be a bug: https://github.com/google/pytype/issues/764. So I added this workaround here, I will add a comment.

salvo/requirements.txt
gyohuangxin: Thanks for this reminder, I have less exprience maintaining a python project, so I didn't notice this. 
It works when I use `>=` instead of `==` both in our dependencies in this PR or original dependencies. It will download the appropriate version automatically. So it's good enough to use '>=' in our original dependencies configuration , it works from python3.6 to python3.10. 

However, I can't understand this, can you give an example:
> This will break CI when the dependency makes an incompatible change, but at least we will know about it soon.

mum4k: Sorry, I should have provided more details. What I meant is - regardless of our configuration, the packages we use will evolve over time and eventually introduce a breaking change. We have two options and which we choose affects how and when we learn about the breaking change:

1. with `>=` our tests in AZP pick up the latest package with the breaking change and will report a test error. This gives us an early warning that we need to update our code due to the upstream breaking change.
2. with `==` our tests in AZP continue using the package version that is compatible with our code. Eventually we either decide to update the dependencies or we will be forced to due to a deprecation of one of the versions we use. This is the moment when we will discover all the breaking changes and since they accumulated - this could become a large and complex PR to get it all in order.

Personally I prefer (1), since it results in smaller (though multiple) updates that are easier to write and review.

gyohuangxin: @mum4k Thanks for providing more details. I can understand completely now, and I also prefer (1) to avoid huge dependencies updates in the future. I will update this PR.

=
```diff
@@ -0,0 +1,2 @@
+Defaulting to user installation because normal site-packages is not writeable
```
mum4k, line 1: Looks like this file may have been created by accident. Or is it needed here?

gyohuangxin: We don't need it here, sorry to add it by mistake. I removed it.

mum4k (approved):

PR #153, opened 2022-06-22
Fix log duplication

xu1zhou:
Add argument to control print behavior of run_command function.
Disable it when unnecessary to print log.

Signed-off-by: xu1zhou <yizhou.xu@intel.com>

### DIFF
--- a/salvo/src/lib/cmd_exec.py
+++ b/salvo/src/lib/cmd_exec.py
@@ -17,7 +17,7 @@ CommandParameters = typing.NamedTuple(
     ])
 
 
-def run_command(cmd: str, parameters: CommandParameters) -> str:
+def run_command(cmd: str, parameters: CommandParameters, debug=True) -> str:
   """Run the specified command returning its output to the caller.
 
   Args:
@@ -61,8 +61,8 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
 
     if isinstance(output, bytes):
       output = output.decode('utf-8').strip()
-
-    log.debug(f"Returning output: [{output}]")
+    if debug:
+      log.debug(f"Returning output: [{output}]")
 
   return output
 
--- a/salvo/src/lib/cmd_exec.py
+++ b/salvo/src/lib/cmd_exec.py
@@ -27,6 +27,7 @@ def run_command(cmd: str, parameters: CommandParameters, debug=True) -> str:
         where the command is to be executed.  Other parameters supported
         by the subprocess module will be added as they become necessary for
         execution.
+      debug: Argument to control default log output print behavior.
 
   Returns:
       The output produced by the command
--- a/salvo/src/lib/builder/base_builder.py
+++ b/salvo/src/lib/builder/base_builder.py
@@ -71,7 +71,7 @@ class BaseBuilder():
 
     cmd_params = cmd_exec.CommandParameters(cwd=self._build_dir)
     cmd = "bazel clean"
-    output = cmd_exec.run_command(cmd, cmd_params)
+    output = cmd_exec.run_command(cmd, cmd_params, False)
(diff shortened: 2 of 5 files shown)


salvo/src/lib/cmd_exec.py
```diff
@@ -17,7 +17,7 @@
     ])
 
 
-def run_command(cmd: str, parameters: CommandParameters) -> str:
+def run_command(cmd: str, parameters: CommandParameters, debug=True) -> str:
```
gyohuangxin, line 20: The new parameter will break the unit tests, please revise their unit tests cases.

salvo/src/lib/cmd_exec.py
```diff
@@ -61,8 +61,8 @@ def run_command(cmd: str, parameters: CommandParameters) -> str:
 
     if isinstance(output, bytes):
       output = output.decode('utf-8').strip()
-
-    log.debug(f"Returning output: [{output}]")
+    if debug:
+      log.debug(f"Returning output: [{output}]")
```
gyohuangxin, line 65: Can we add a DEBUG flag in log message to indicate it's a debug information?

gyohuangxin (changes requested):

xu1zhou:
it seems something happend to CI test. I think it's not due to my modification cause the failed. @gyohuangxin  Do you have any clue?

gyohuangxin:
@xu1zhou Hi, don't worry about the CircleCI test failure, it often fails due to the OOM. Please focus on the test results of Azure pipleines, they work fine.
image

xu1zhou:
> @xu1zhou Hi, don't worry about the CircleCI test failure, it often fails due to the OOM. Please focus on the test results of Azure pipleines, they work fine. image

Thanks, will check reason for that

salvo/src/lib/cmd_exec.py
gyohuangxin: @xu1zhou Adding a new parameter here will break the unit tests and you have to resolve many mock errors. Because this function is a lower level function used in many place, any small change may cause many unexpected issues, struggling with this may be a waste of time. 

Salvo is using `logging` package to manage logs, so you can use its functions such as [setlevel](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel) to change the behaviors without adding any parameter.

xu1zhou:
Hi @gyohuangxin I think pipeline test has passed this time.

gyohuangxin (approved):
LGTM, thanks for your improvement.

salvo/src/lib/cmd_exec.py
mum4k: +1 on this comment. Adding a logging specific parameters makes the API more complicated and we do seem to have a viable alternative (setting the logging level).

Can we attempt a solution that will modify the logging level to compare? If you want, we can keep this PR around just in case we decide to go back to this and send a new PR that will set the logging level.

mum4k (changes requested):

gyohuangxin:
@xu1zhou Yes, setting logging level will affect all the functions which use `run_command`. 
The default log level is `debug`, so how about to set the logs of `run_command` to `trace` level? For example, `log.trace(f"Returning output: [{output}]")`
If someone want to see this log, he/she can set the log level to `trace` without change many functions' parameter value to `True`.

gyohuangxin:
@xu1zhou Do you have any updates?

xu1zhou:
> @xu1zhou Do you have any updates?

Sorry, busy with some other thing these day. I would have a update this week.

gyohuangxin:
@xu1zhou Don't worry about this, thanks.

mum4k:
Closing for inactivity. Feel free to reopen this when ready.

PR #165, opened 2022-11-08
Syncing configuration files from Nighthawk and Envoy (11/08/2022)

gyohuangxin:


mum4k:
@gyohuangxin it looks like our CI was broken sometime between the last PR and this one. It fails both at head and at this PR with this error:

```
E           docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', PermissionError(13, 'Permission denied'))
```

If you don't have the time to investigate, I can look into it next week. We probably don't want to submit this PR without being able to verify it still works, so we should fix the tests first.

gyohuangxin:
@mum4k I've had CI broken before, but I can't remember the reason at that time, it worked after retest. I agree that we need to fix it at first, I can speed some time to investigate it this week.

gyohuangxin:
@mum4k Please merge it, thanks!

mum4k (approved):

PR #181, opened 2022-12-05
Use a Python virtual environment for Salvo.

mum4k:
Salvo now uses a virtual environment which will provide better package separation on the local machine.

Done here:
- removed the `install_deps*` scripts. The scripts required users to execute a shell script as root which creates trust issues. Most Python packages can be installed at user level privileges into the virtual environment. The instructions on how to install system level packages have been moved to the readme file where users have full visibility into what they are running as root.
- changed the "building Salvo" command in the readme to our `ci/do_ci.sh` script which sets up the virtual environment.
- Moved the `python-apt` package from `requirements.txt` into the system level `apt` command line to avoid having ubuntu specific package versions in `requirements.txt`.
- Removed `pygobject` from `requirements.txt`, because it requires a system level package to be installed and doesn't seem to be required for Salvo.
- added `pytype` into requirements since it is needed for the type checking CI phase.
- sorted `requirements.txt` lexicographically.


mum4k:
@fei-deng please review and assign back to me once done.

gyohuangxin (approved):
Good idea to use a virtual environment, thanks!

PR #195, opened 2023-03-14
Syncing configuration files from Nighthawk and Envoy (03/14/2023)

gyohuangxin:


gyohuangxin:
Investigating the failure...

phlax:
hi @gyohuangxin i see the docker-related CI is failing, ive made quite a few changes to this recently

im wondering if you are using docker inside the build container ?

there are some other possibilities i can think of also, lmk if i can help

gyohuangxin:
@phlax Thanks for your help. Yes, I think so. I can run the same test on my own machine, but it failed in the AZP pipeline.

phlax:
@gyohuangxin trying to understand relationship with nighthawk - but see https://github.com/envoyproxy/nighthawk/pull/977 for related fix

gyohuangxin:
@phlax The CI pipeline of envoy-perf is similar with nighthawk, I will try it.

phlax:
just flagging that the reason it is now disabled by default is that it breaks the security model of running inside a container - anything running with the default user inside the container can trivially get root on the host

mum4k (approved):
Thank you both. I am going to rework the CI pipeline here according to [this design](https://docs.google.com/document/d/1auXzV-AEXgMzbtdG06XlZ2d9X5l_XadMheA9h51E7yc/edit#bookmark=id.k5jg61c9551x), so this will go away shortly.

PR #197, opened 2023-03-15
Salvo initial pipeline.

mum4k:
An initial AZP pipeline for the MVP of Salvo. Checks out `nighthawk` and `envoy` repositories, builds and publishes the binaries.

Next step will be to integrate Packer, so that VM images are built.

Pipeline result: https://dev.azure.com/cncf/envoy/_build/results?buildId=131099&view=results

PR #198, opened 2023-03-21
Packer scripts that build AMI for Salvo's component VMs.

mum4k:
These scripts will be executed from the AZP pipeline added in https://github.com/envoyproxy/envoy-perf/pull/197.

Tested on AWS.


gyohuangxin (approved):
LGTM, thanks!

fei-deng (approved):
LGTM!

PR #209, opened 2023-05-03
Salvo MVP pipeline fixes.

mum4k:
- the location of the Envoy binary has changed (new `x64` subdirectory).
- the VPC and Security Group IDs changed when I moved Salvo to its own region.


eric846 (approved):
LGTM

PR #239, opened 2023-11-01
Syncing configuration files from Nighthawk and Envoy (11/01/2023)

gyohuangxin:
Relates to #240

### DIFF
--- a/ci/run_envoy_docker.sh
+++ b/ci/run_envoy_docker.sh
@@ -127,6 +127,7 @@ docker run --rm \
        -e CI_TARGET_BRANCH \
        -e DOCKERHUB_USERNAME \
        -e DOCKERHUB_PASSWORD \
+       -e ENVOY_DOCKER_SAVE_IMAGE \
        -e ENVOY_STDLIB \
        -e BUILD_REASON \
        -e BAZEL_REMOTE_INSTANCE \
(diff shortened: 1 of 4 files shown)


gyohuangxin:
The error logs are:
```
Creating a new Python virtual environment for Salvo in salvo_venv/venv.
tools/python_virtualenv.sh: line 12: virtualenv: command not found
##[error]Bash exited with code '127'.
```

It seems that virtualenv is not installed in new docker image, I will try to fix it.

ci/run_envoy_docker.sh
```diff
@@ -59,7 +59,7 @@ else
           && usermod -a -G pcap envoybuild \
           && chown envoybuild:envoygroup ${BUILD_DIR_MOUNT_DEST} `# Salvo unique` \
           && chown envoybuild /proc/self/fd/2 \
-          && apt-get update && apt -y install libcairo2-dev `# Salvo unique` \
+          && apt-get update && apt -y install libcairo2-dev python3-virtualenv `# Salvo unique` \
```
mum4k, line 62: Instead of adding support for on-host Python back into the image, we should work on making the Salvo build hermetic.

Here is an example PR where this was done in the Nighthawk codebase:
https://github.com/envoyproxy/nighthawk/pull/1033

Is this something you would be able to take on? We can submit the current PR as a workaround, but we should file an issue and work towards hermetic builds, as the current setup can cause unexpected version changes.

gyohuangxin: @mum4k  Thanks for your comments! Of course, I'd like to make Salvo build hermetic, I'll create and take on the issue.

mum4k (approved):

ci/run_envoy_docker.sh
mum4k: Thank you for taking this on!

PR #241, opened 2023-11-08
Update links to Salvo documents.

mum4k:
The documents were migrated from Google to MS cloud.

Removed links to ASANA which no longer exist.


gyohuangxin (approved):
LGTM

